PR #18, opened 2019-08-06
comment out line making lots of noise in FSOI ncdiag implementation

will-mccarty:


mathomp4:
@will-mccarty I think @gmao-jstassi might be out for the month but, because I forgot to remove myself, I'm still a member of @GEOS-ESM/adas-gatekeepers. 

Is this a crucial issue that needs fixing immediately, or can you wait until @gmao-jstassi returns so it's more "official"?

mathomp4:
Also, note for Joe (and I'll ping @tclune), since we are currently just using `master` for `GMAO_Shared`, we'd need a new release to officially use this so we can advance `Externals.cfg`.

will-mccarty:
This was one small test that'll go along with some larger changes in the GSI code in another pull request.

will-mccarty:
No rush on it

PR #32, opened 2019-09-03
Prescribing Dynamic Vegetation Parameters for CATCHCN

smahanam:
**PURPOSE:** 

The CatchmentCN (CATCHCN) model may replace the catchment (CATCH) model in the next S2S forecasting system. CATCHCN is significantly slower than CATCH mainly due to the computer time of Carbon and Nitrogen module (CLM). Thus, we decided to bypass the CLM in the S2S implementation, yet capture the variability of dynamic vegetation parameters of CLM, by implementing a system that prescribes daily climatologies of LAI and SAI from CLM and uses a pre-determined time scale parameters.
The script and the program in this branch works as the pre-processor to set up CTACHCN prescribe DYNVEG parameters for S2S. 

**NOTE** 
I am aware that S2S is run on NAS and there is a hard-wired path to a directory in discover in the program. I have tested the software which works the way we intend it to. For now, I just want to tag this with the develop branch (Note that there is no develop branch, so I had to branch off master) for safe keeping until we finalize S2S implementation.

### DIFF
--- a/GEOS_Util/pre/CMakeLists.txt
+++ b/GEOS_Util/pre/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(NSIDC-OSTIA_SST-ICE_blend)
+add_subdirectory(DYNVEG)
--- a/GEOS_Util/pre/DYNVEG/CMakeLists.txt
+++ b/GEOS_Util/pre/DYNVEG/CMakeLists.txt
@@ -1,0 +1,12 @@
+# No library
+esma_set_this(OVERRIDE raster)
+set (srcs regrid_DYNVEG.F90)
+
+include_directories(${INC_ESMF})
+include_directories(${INC_NETCDF})
+
+esma_add_library(${this} SRCS ${srcs} DEPENDENCIES MAPL_Base)
+
+ecbuild_add_executable (TARGET regrid_DYNVEG.x SOURCES regrid_DYNVEG.F90  LIBS MAPL_Base ${this})
+
+install(PROGRAMS regrid_DYNVEG.csh DESTINATION bin)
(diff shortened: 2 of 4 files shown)


JustinPerket:
Is this related to https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61 ?

tclune:
@JustinPerket Can you or someone appropriate please review this PR so that it can be closed.   Also if it could be annotated as zero-diff it might be useful in future forensics.

smahanam:
Apologies if my description is not clear.
These modifications are for CATCHCN which itself is not the default model. So it is zero-diff.
This is a pre-processor sits outside the model physics. One would run to set up a PRESCRIBE_DVG experiment for CATCHCN from the experiment directory. The pre-processor would create some input files for user grid/tiles.

tclune:
@smahanam You may have been clear - I was just trying to get this cleared through the system and could not immediately spot any statement  in a quick scan.    Still would like to get one other relevant scientist to concur with a review before I merge.

JustinPerket:
No objections. Will test do a quick test of functionality. but it looks like it requires a file from
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61/ :

https://github.com/GEOS-ESM/GMAO_Shared/blob/8c3ae56b5c3c76a2b0e14456d7a0e9667a910a60/GEOS_Util/pre/DYNVEG/regrid_DYNVEG.csh#L36

That pull request is on hold for the moment. ```GEOS_LandGridComp.rc``` should be put on it's own pull request. Additionally, this pull request is also needed for functionality:
https://github.com/GEOS-ESM/GEOSgcm_App/pull/39

This is all related to this issue, which I will update to centralize that discussion:
https://github.com/GEOS-ESM/GEOSgcm_App/issues/30

smahanam:
If it helps, I can send a separate PR for this particular file GEOS_LandGridComp.rc. It is where land physics parameters sits - so everything is linked to that file.

JustinPerket:
As suspected, fails with error 
> grep: RC/GEOS_LandGridComp.rc: No such file or directory

@smahanam, yes, let's get GEOS_LandGridComp.rc in and functional GEOS-ESM/GEOSgcm_App#30

fwzeng:
Yes, S2S forecast group showed an interest in replacing the current CATCH with CATCHCN. We designed this efficient system particularly for the S2S because fully coupled CATCHCN model requires about 7 times more computer time than CATCH.

PR #61, opened 2019-12-05
Feature/sm land physics

smahanam:
This PR is related to the updates in[ PR 151](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/151)

Preprocessors for Issues [146](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/146) and [147](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/147) are here.


GEOS_Util/pre/DYNVEG/regrid_DYNVEG.csh
```diff
@@ -0,0 +1,111 @@
...
+setenv YEARB $YEARB
+setenv YEARE $YEARE
+setenv ACTUAL_DV $ACTUAL_DV
+
+limit stacksize unlimited
+
+cd ${EXPDIR}/VEGDATA/scratch
+
+##mpirun -map-by core --mca btl ^vader -np 56 $GEOSBIN/regrid_DYNVEG.x
+mpirun -np 56 $GEOSBIN/regrid_DYNVEG.x
```
mathomp4: Query for @smahanam: will this code ever run for anyone *but* LDAS? The reason I ask is that `mpirun` is not the correct MPI executable for some MPI stacks. MPT uses `mpiexec_mpt` and MVAPICH uses either `srun` or `mpirun_rsh`. If you want to be agnostic, it would be better to use `$GEOSBIN/esma_mpirun` like how `regrid.pl` does.

(Also, you've hardcoded this for discover in the #SBATCH above, but my guess is the needed BCs don't exist at NAS?)

JustinPerket:
> 1. The changes seem touch on two unrelated topics:  (i) the revised regridding (weminIN and weminOUT...) and (ii) a newer functionality that handles dynamically varying vegetation and albedo inputs.  Do the latter depend on other PRs?

Agreed about WEmin. There is not yet a GMAO_Shared PR, but there's a branch for the regridding updates relating to WEmin.
Issue: https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/116
Branch: https://github.com/GEOS-ESM/GMAO_Shared/compare/feature/jstassi/added-wemin-wemout-options

If these are updates for WEmin, I suggest they get merged into that branch. But it looks like they're an older version? And should just be removed from this branch to avoid confusion.

> 2. If I'm reading this correctly, the PR is to merge into "master".  Is this already on develop? 
Not all repositories have enough activity to need a separate develop branch. It looks like GMAO_Shared is one of those.

tclune:
Is this PR still desired?   It is over a year old.     Should we expedite or delete?

smahanam:
OK I'll close the PR for now, shall keep the branch though.

PR #69, opened 2020-01-23
Shared module for lightning flash rate computation

christophkeller:
Consolidate the calculation of lightning flash rates (LFR). This module was created by Megan Damon to bring together a suite of LFR parameterizations, which are called from various parts of the model (e.g., MOIST, chemistry).
This PR has no dependencies and should result in zero-diff.


christophkeller:
Any chance we can move forward with this pull request? This pull request needs to be merged before two other pull requests can be moved forward ( [PR187 in GEOSgcm_GridComp](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/187) and [PR46 in GEOSchem_GridComp](https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/46) ). I would like to start adding additional updates to GEOSCHEMchem but would like this PR to be completed first.

tclune:
I think the problem is simply that there is no real gatekeeper for GMAO shared.   Anyone can click the button so ... no one does.

christophkeller:
Perfect, thanks!

PR #76, opened 2020-02-11
Revised help message for WEMIN in regrid.pl [skip ci]

sdrabenh:


bena-nasa (dismissed):

lltakacs:
@gmao-rreichle 
Rolf, I have some questions related to the wording you supplied related to the WEmin selection in regrid.pl:

1) Since the user has already specified the boundary condition area they want to use, then the default choice for WEmin is already known.  A question isn't really needed.

Thus, the real question is:  Does the user want to supply their own value?

If the user supplies their own value, do they also have to create a new boundary area consistent with their value, or is Icarus-NLv3 still appropriate?  

If they need a new boundary area, then REGRID.pl already has a question asking for alternative user-supplied boundary locations.  We can then ask what value for WEmin the user wants to use with their alternative boundary area.

If they can use Icarus-NLv3, then we should simply ask for the value to use.  


2) Your comments specify associations with MERRA, MERRA-2, and MERRA-Land, where you have linked MERRA with the newer WEmin value of 13.  Is this accurate? 

To me, the new wording begs more questions and uncertainty than the original version.

gmao-rreichle:
Larry, @lltakacs 
Let me try to answer your questions here.   If I'm not successful, it may be easier to talk in person.

> Since the user has already specified the boundary condition area they want to use, then the default > choice for WEmin is already known. A question isn't really needed.

The problem is that there is not a firm link between a particular set of bcs and the WEMIN value that is used in the code.   We would have to create some sort of look-up table to come up with the WEMIN values that were *typically* used with a given set of bcs.  This is, in a nutshell, what the help message is trying to accomplish.  But there's no guarantee that a user did ("in") or will want to ("out") use the WEMIN value that is *typically* used with a given set of bcs. 

> If the user supplies their own value, do they also have to create a new boundary area consistent 
> with their value, or is Icarus-NLv3 still appropriate?

Rephrasing the above, WEMIN is a runtime model input parameter, and it is entirely decoupled from the generation of the  bcs.  The issue at hand is spin-up.  If the user changes the WEMIN value from what was used in the run that generated the source ("in") restart, then the snow mass and depth values in the restart must be adjusted accordingly, or the snow density gets messed up, which then would mess up the validity of the spun-up restart. 
 
> If they need a new boundary area, then REGRID.pl already has a question asking for alternative 
> user-supplied boundary locations. We can then ask what value for WEmin the user wants to use 
> with their alternative boundary area.

There's no need for a new boundary area.  The source ("in") boundary area must of course match what was used in the run that generated the bcs.   The target ("out") boundary area is whatever the user wants it to be.   Likewise for WEMIN.  And, again, there's no general way of detecting from the source  restart what WEMIN value was used in the run that generated the source restart.  

> 2. Your comments specify associations with MERRA, MERRA-2, and MERRA-Land, where you 
> have linked MERRA with the newer WEmin value of 13. Is this accurate?

The original MERRA indeed used the value of 13.  The bad value of 26 was picked by non-land developers when they tried to get Fortuna to behave.  I was not part of the process at the time, so my knowledge of the history is limited.  But I'm sure the original MERRA used WEMIN=13.

> To me, the new wording begs more questions and uncertainty than the original version.

That, of course, is not what I intended.  But the original wording clearly stumped Ricardo, who emailed us "How am I supposed to answer this?" when he first came across the WEMIN query.

I'm happy to brainstorm more about better wording.  Maybe we can also come up with educated guesses and present them to the user (if that isn't in the script already, not sure).  But there is simply no general way for a script to always figure out the correct WEMIN values.

- Rolf

tclune:
This PR is nearly a year old.     Is it still desired?   Should it be expedited or deleted?

mathomp4:
I'm labeling as trivial because, well, it is a documentation sort of edit. But I suppose it's up to @GEOS-ESM/ldas-gatekeepers to approve wording.

gmao-rreichle:
Thanks, @tclune, for reminding us that this is still pending.  
Yes, I still think we want to make this minor change.  
I tweaked the message a bit more after re-reading the conversation, to address the comments by @lltakacs .
As @mathomp4 said, this is really just documentation, so trivially 0-diff.
Unless there are more objections, I would suggest proceeding with the merger, before we forget about the PR again.
cc: @biljanaorescanin @rdkoster @weiyuan-jiang @sdrabenh

mathomp4 (approved):

PR #81, opened 2020-03-12
Feature/wjiang/#149 add write eta

weiyuan-jiang:
1) divide the module m_topo_remap into two modules. The module shared_topo_remap is independent of any other part of GMAO_hermes and used by FV core
2)write_eta.F90 is added to prepare for a config file ( or rc file) which eventually will replace m_set_eta module


CMakeLists.txt
```diff
@@ -1,36 +1,44 @@
...
-  GMAO_pilgrim
-  GMAO_etc
-  GEOS_Shared
-  GMAO_hermes
-  GEOS_Util 
-  LANL_Shared
-  GMAO_perllib
-  GMAO_transf
-  GMAO_stoch
+if ( DEFINED HERMES_LIGHT )
```
tclune: 2 problems here:    The big one is that HERMES_LIGHT should always be defined.   It should be added via a CMake OPTION.   

The tiny issue, is that I think it would be better to just always include the two directories and then IF (NOT HERMES_LIGHT)  add in the others.   (I.e., no ELSE clause)

GMAO_hermes/CMakeLists.txt
```diff
@@ -1,78 +1,94 @@
...
+set (CMAKE_Fortran_FLAGS_RELEASE "-O3 ${common_Fortran_flags} ${ALIGNCOM}")
 
-set (srcs
-  m_ana2dyn.F m_const.f90 m_dyn.f90 m_dyn2dyn.f90 m_dynp.F90
-  m_insitu.F m_interp.F90 m_lp.F m_maph.F90 m_maph_pert.F90 m_mapz.F m_mapz_pert.F90
-  m_qsat.F m_remap.f90 m_set_eta.F90 m_vdc2vdc.F90 progsig.f90 m_topo_remap.F90
-  m_interpack.F m_interpack_ad.F m_ec_set_eta.F90 m_ecdyn.f90 m_tick.f90
-  m_dyn_util.F90 m_spline.f90
-  )
+if ( NOT DEFINED HERMES_LIGHT )
```
tclune: As above,  "DEFINED" should not be here.

GMAO_hermes/unused_dummy.H
```diff
@@ -0,0 +1,13 @@
+! The following macro causes a variable to appear to be "used"
```
tclune: I'm a bit confused that this file was needed.  The directory already had _UNUSED_DUMMY calls, so ... how were they satisfied?

GMAO_hermes/write_eta.F90
```diff
@@ -0,0 +1,118 @@
...
+     call set_ncep72()
+  endif
+
+  if (levels == 0 ) stop "-l is a must"
+
+  allocate(ak(levels+1),bk(levels+1))
+
+  call set_eta(levels, ks, ptop, pint, ak, bk)
+
+  open(10, file="eta.rc", action="write", FORM="FORMATTED")
```
tclune: Please use the NEWUNIT option in open rather than hardcoded units for files.   Granted, in this case it could never break anything, but we want to also encourage good practices elsewhere.

More importantly there should be a status="new" so that we cannot accidentally overwrite a file.

GMAO_hermes/write_eta.F90
```diff
@@ -0,0 +1,118 @@
...
+      endif
+      if(abs(bk_tmp-bk(k)) > 0.0d0) then
+         print*, k
+         stop " bk is not consistent"
+      endif
+      if (bk_tmp > 0.0d0 .and. ks_tmp == -1 ) then
+           ks_tmp = k-2
+           if (ks_tmp /= ks) then
+              print*, "ks : ", ks_tmp
+              stop "transition to pure pressure level is not correct"
```
tclune: change to  error stop

GMAO_hermes/write_eta.F90
```diff
@@ -0,0 +1,118 @@
...
+    if (status < 0) exit
+    if(index(line,"ak-bk") == 0)cycle
+
+    ks_tmp = -1
+    do k = 1, levels + 1
+      read(10,'(A)', IOSTAT = status) line
+      read(line,*) ak_tmp, bk_tmp
+      if(abs(ak_tmp-ak(k)) > 0.0d0) then
+         print*, k
+         stop " ak is not consistent"
```
tclune: change to error stop

GMAO_hermes/unused_dummy.H
weiyuan-jiang: With HERMES light, there is no any include_directory, so it cannot find _UNUSED_DUMMY calls

GMAO_hermes/write_eta.F90
weiyuan-jiang: I think I should use the esmf_configloadFile to read the data . That is the way how this file will be used.

weiyuan-jiang: give up. That creates dependency on esmf.

weiyuan-jiang: All your comments are addresses.

weiyuan-jiang:
The code is re-organized @tclune

PR #84, opened 2020-03-25
Fixes for CMake using F2Py and LaTeX

mathomp4:
This update fixes some issues where systems do not have
either ImageMagick or F2Py. If F2Py is not found, then F2Py
targets are not built.

This should be taken in concert with:

GEOS-ESM/ESMA_cmake#62
GEOS-ESM/GEOSchem_GridComp#68
GEOS-ESM/GMAO_Shared#84
GEOS-ESM/UMD_Etc#11

That said, if the ESMA_cmake PR/change is not taken, the model will still build, but no f2py will run (as it's missing the `F2PY_FOUND` variable).


mathomp4:
Adding @sdrabenh as a reviewer mainly to keep him aware of the multi-repo-ness of this change.

PR #90, opened 2020-04-04
The regrid.pl script was updated for SLES-12

gmao-jstassi:
This change is in conjunction with updates in mk_Restarts in the @GEOSgcm_GridComp component. I ran tests that were successful on both SLES-11 and SLES-12.


sdrabenh:
@gmao-jstassi should the hard coded #SBATCH directives (line ~2382) in regrid.j be changed depending on whether running at NAS or NCCS?

mathomp4:
> @gmao-jstassi should the hard coded #SBATCH directives (line ~2382) in regrid.j be changed depending on whether running at NAS or NCCS?

@sdrabenh I don't think we really support `regrid.pl` at NAS. It doesn't have access to MERRA and MERRA2 there so a good chunk of functionality is gone. I guess we could support non-MERRA/MERRA2 regridding at some point though we'd need to fix a lot.

I suppose we maybe should have a detection to fail if not at NCCS. (And, I guess, change all the "PBS" references in it to SLURM.)

sdrabenh:
> > @gmao-jstassi should the hard coded #SBATCH directives (line ~2382) in regrid.j be changed depending on whether running at NAS or NCCS?
> 
> @sdrabenh I don't think we really support `regrid.pl` at NAS. It doesn't have access to MERRA and MERRA2 there so a good chunk of functionality is gone. I guess we could support non-MERRA/MERRA2 regridding at some point though we'd need to fix a lot.
> 
> I suppose we maybe should have a detection to fail if not at NCCS. (And, I guess, change all the "PBS" references in it to SLURM.)

Interesting, I wasn't aware that we don't support regridding at NAS. In that case, I guess my comments are a moot point. Still, it would seem a good goal in the long term even if they aren't necessary now.

PR #107, opened 2020-07-09
Use r4i4 ncep_sp in gmao_transf

bena-nasa:
Use this r4i4 version of ncesp_sp in gmao_transf. This is needed so that the GEOSgcm.x will link to this version allowing the JCAP functionality to work again in mkiau gridcomp as that needs the r4 version. Note this is non-zero diff if using stochastic phyiscs (SPPT: 1 and SKEP: 1). However, I would be suspicious of any answer from that as if you turn on that option with debugging it fails in the stochastic physics with an out of bounds error.


mathomp4:
I added the non-0-diff label, though...I guess it's hard to label "Allows code to work" as being 0-diff or non-0-diff...so I'll add 0-diff as well. Hmm.

PR #110, opened 2020-08-10
Fixes to allow Intel 19.1.2 to use CICE

mathomp4:
For some reason, Intel really wasn't happy with opening a file everywhere.

### DIFF
--- a/LANL_Shared/LANL_cice/source/ice_grid.F90
+++ b/LANL_Shared/LANL_cice/source/ice_grid.F90
@@ -376,8 +376,10 @@
 
          if (trim(grid_format) == 'nc') then
 
-            call ice_open_nc(grid_file,fid_grid)
-            call ice_open_nc(kmt_file,fid_kmt)
+            if (my_task == master_task) then
+               call ice_open_nc(grid_file,fid_grid)
+               call ice_open_nc(kmt_file,fid_kmt)
+            endif
 
             fieldname='ulat'
             call ice_read_global_nc(fid_grid,1,fieldname,work_g1,.true.)
@@ -391,8 +393,10 @@
 
          else
 
-            call ice_open(nu_grid,grid_file,64) ! ULAT
-            call ice_open(nu_kmt, kmt_file, 32) ! KMT
+            if (my_task == master_task) then
+               call ice_open(nu_grid,grid_file,64) ! ULAT
+               call ice_open(nu_kmt, kmt_file, 32) ! KMT
+            endif
 
             call ice_read_global(nu_grid,1,work_g1,'rda8',.true.)  ! ULAT
             call ice_read_global(nu_kmt, 1,work_g2,'ida4',.true.)  ! KMT
@@ -406,7 +410,9 @@
 
       elseif (trim(grid_type) == 'panarctic') then
 
-         call ice_open(nu_grid,grid_file,64) ! ULAT, KMT
+         if (my_task == master_task) then
+            call ice_open(nu_grid,grid_file,64) ! ULAT, KMT
+         endif
 
          call ice_read_global(nu_grid,1,work_g2,'ida8',.true.)  ! KMT
          call ice_read_global(nu_grid,2,work_g1,'rda8',.true.)  ! ULAT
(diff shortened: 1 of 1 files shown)


mathomp4:
I'll make this "ready for review", but I'll wait on some tests from @yvikhlya just to make sure fixing things for Intel 19.1.2 doesn't break things for Intel 18.0.5!

mathomp4:
> Should it be tested with old compiler as well?

@zhaobin74 Yes. Probably a good idea. Main issue is that Intel 19.1 is non-zero-diff to Intel 18, so "works" is about all I can say for Intel 19. But if it's zero-diff with Intel 18, then at least we know this shouldn't be a cause of any change with Intel 19.

zhaobin74:
@mathomp4, what I can do is to bring this change back to CVS version and do a zero-diff test with the old compiler (18.x). I expect to see zero-diff. If so, I'll approve. Will it be good enough?
> > Should it be tested with old compiler as well?
> 
> @zhaobin74 Yes. Probably a good idea. Main issue is that Intel 19.1 is non-zero-diff to Intel 18, so "works" is about all I can say for Intel 19. But if it's zero-diff with Intel 18, then at least we know this shouldn't be a cause of any change with Intel 19.

mathomp4:
> @mathomp4, what I can do is to bring this change back to CVS version and do a zero-diff test with the old compiler (18.x). I expect to see zero-diff. If so, I'll approve. Will it be good enough?

Sounds good to me. I'm at the stage where I'm happy I can *run* MOM5/MOM6. I still rely on you all to make sure it's running correctly 😄

zhaobin74:
@mathomp4, I am a bit confused about this fix. If you look at `ice_open` and `ice_open_nc` subroutine, there are already protections with `if (my_task == master_task)`  block. Why adding another layer solves problem?

mathomp4:
> @mathomp4, I am a bit confused about this fix. If you look at `ice_open` and `ice_open_nc` subroutine, there are already protections with `if (my_task == master_task)` block. Why adding another layer solves problem?

@zhaobin74 I agree it is odd. My only thought is that some how calling a routine that does nothing on all processors but master causes...something? I was just trying everything I could to keep Intel 19.1.2 happy

zhaobin74:
> > @mathomp4, I am a bit confused about this fix. If you look at `ice_open` and `ice_open_nc` subroutine, there are already protections with `if (my_task == master_task)` block. Why adding another layer solves problem?
> 
> @zhaobin74 I agree it is odd. My only thought is that some how calling a routine that does nothing on all processors but master causes...something? I was just trying everything I could to keep Intel 19.1.2 happy

Ok, I'll proceed with the zero-diff test. As long as Intel19 is happy, I am fine with that extra layer.

yvikhlya:
> > > @mathomp4, I am a bit confused about this fix. If you look at `ice_open` and `ice_open_nc` subroutine, there are already protections with `if (my_task == master_task)` block. Why adding another layer solves problem?
> > 
> > 
> > @zhaobin74 I agree it is odd. My only thought is that some how calling a routine that does nothing on all processors but master causes...something? I was just trying everything I could to keep Intel 19.1.2 happy
> 
> Ok, I'll proceed with the zero-diff test. As long as Intel19 is happy, I am fine with that extra layer.

mathomp4:
@yvikhlya Is there a reason you closed this?

yvikhlya:
@mathomp4 Oh, I am sorry, I was writing a comment and pressed "close" by accident instead of "cancel". My fault.

mathomp4:
> @mathomp4 Oh, I am sorry, I was writing a comment and pressed "close" by accident instead of "cancel". My fault.

No worries. I thought perhaps you found a better fix! 😄

yvikhlya:
@mathomp4 This fix works with intel-18 and zero diff with 1.1.7.

zhaobin74:
Zero-diff confirmed on CVS tag with Intel 18.

mathomp4:
@zhaobin74 I think because I pushed something, your review was lost. Can you re-review this?

zhaobin74:
@yvikhlya, can you push the cvs tag as well? My tag 'bzhao-s2s3-unstable' has it.

PR #115, opened 2020-08-26
Update GEOS_Utilities to match ADAS

mathomp4:
The `GEOSadas-5_27_0` version of this file has slightly different code, `nint()` instead of `int()`. This brings us inline. 

Marking as non-zero-diff as it might be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update `GEOS_Utilities.F90` to match `GEOSadas-5_27_0` (#115)
+
 ### Fixed
 
 ### Removed
--- a/GEOS_Shared/GEOS_Utilities.F90
+++ b/GEOS_Shared/GEOS_Utilities.F90
@@ -535,7 +535,7 @@
        end if
 
        TI = (TI - TMINTBL)*DEGSUBS+1
-       IT = int(TI)
+       IT = nint(TI)
 
        if(URAMP==TMIX) then
           DQ    = ESTBLX(IT+1) - ESTBLX(IT)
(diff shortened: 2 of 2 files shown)


mathomp4:
Tagging @rtodling to make sure I got this right, but the diff from CVS looks the same:
```diff
Index: GEOS_Utilities.F90
===================================================================
RCS file: /cvsroot/esma/esma/src/Shared/GEOS_Shared/GEOS_Utilities.F90,v
retrieving revision 1.16.326.2
retrieving revision 1.16.326.1
diff -b -u -p -r1.16.326.2 -r1.16.326.1
--- GEOS_Utilities.F90	20 May 2020 15:30:30 -0000	1.16.326.2
+++ GEOS_Utilities.F90	24 Mar 2016 19:41:09 -0000	1.16.326.1
@@ -535,7 +535,7 @@
        end if

        TI = (TI - TMINTBL)*DEGSUBS+1
-       IT = nint(TI)
+       IT = int(TI)

        if(URAMP==TMIX) then
           DQ    = ESTBLX(IT+1) - ESTBLX(IT)
@@ -724,7 +724,7 @@
          end if

          TT = (TI - TMINTBL)*DEGSUBS+1
-         IT = nint(TT)
+         IT = int(TT)

          if(URAMP==TMIX) then
             DQQ =  ESTBLX(IT+1) - ESTBLX(IT)
```

tclune:
@mathomp4 I thought we determined that the existing logic was safe.    Failures, or at least some of them, were tracked back to uninitialized values in MOIST and my own work showed the indices cannot go out of bounds (for non-NaN values).

mathomp4:
> @mathomp4 I thought we determined that the existing logic was safe. Failures, or at least some of them, were tracked back to uninitialized values in MOIST and my own work showed the indices cannot go out of bounds (for non-NaN values).

@tclune Yes, I think that is the case, and that this is probably a no-op, but it has been adopted in GEOSadas-5_27_0, so I guess I figure we might as well be consistent. It'll come in at some point when the ADAS moves to git.

sdrabenh (approved):

PR #118, opened 2020-09-14
Update the changelog to present day

mathomp4:
Make the Changelog useful!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,15 @@
 # Order is important; the last matching pattern takes the most
 # precedence.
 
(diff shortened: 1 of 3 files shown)


tclune (dismissed):
Thanks @mathomp4.

Maybe we should add the SI team as CODEOWNERS for this layer to ensure that the CHANGELOG is maintained?

mathomp4:
> Thanks @mathomp4.
> 
> Maybe we should add the SI team as CODEOWNERS for this layer to ensure that the CHANGELOG is maintained?

Okay, Let me add that...

tclune (approved):

PR #126, opened 2020-10-02
Feature/san akel/aoil v2.0 dev

sanAkel:
Updates in module `atmOcnIntlayer` :

resolve https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/356


sanAkel:
@tclune, I recall you said something about the change log. But I don't know what to add ... even in a `PR`?

If yourself or @mathomp4 give some clues, I will fill in what is needed. 
Thank you!

GEOS_Shared/atmOceanIntLayer.F90
```diff
@@ -3,53 +3,330 @@ module atmOcnIntlayer
 ! !USES:
 
 use MAPL
+use GEOS_UtilsMod ! for GEOS_DQSAT and GEOS_QSAT
 
 implicit none
-private SIMPLE_SW_ABS, AOIL_SST
-public  ALBSEA, COOL_SKIN, SKIN_SST
+
+private SIMPLE_SW_ABS,      &
```
tclune, line 10: A default PRIVATE clause is preferred to an explicit list.     Otherwise if someone inadvertently declares a module variable further down it will default to PUBLIC.      I don't immediately recall if you can "confirm" the default PRIVATE with explicit, but I think you can.     Otherwise you can always mention the explicit list of private things in a comment.

tclune (changes requested):
Since there is only the one small change requested,  and GitHub showed your branch was initially out-of-date, it is probably best to make the edit directly on GitHub.

tclune:
I have added e template ChangeLog.md to the PR so satisfy our new requirement.     Would not hurt if @sanAkel could edit to add content.

mathomp4:
@sanAkel I think @tclune and I figured this out. You hadn't updated/merged in the latest `main` into your branch, and thus it didn't have the `CHANGELOG.md` that was added after you branched off.

But, the branch was updated by GitHub (clicked a button) and your branch should now have a `CHANGELOG.md` file. That is what requires an edit.

GEOS_Shared/atmOceanIntLayer.F90
sanAkel: @tclune thank you! This is done.

sanAkel:
> @sanAkel I think @tclune and I figured this out. You hadn't updated/merged in the latest `main` into your branch, and thus it didn't have the `CHANGELOG.md` that was added after you branched off.
> 
> But, the branch was updated by GitHub (clicked a button) and your branch should now have a `CHANGELOG.md` file. That is what requires an edit.

Updated `CHANGELOG.md` Thanks!

tclune (approved):
Looks good.  Thanks.

sdrabenh (approved):

PR #132, opened 2020-10-15
Add logger for NaN detection in QSAT calls

mathomp4:
This PR is to add logger calls (at a WARNING level) for the QSAT and DQSAT codes. Recently, a case was found where a NaN was passed down into GEOS_QSAT and this led to a crash. With these loggers, the user would at least get a warning that a NaN was passed before the model inevitably crashes.

Example output is:
```
            QSAT: WARNING:  QSAT0: TL contains NaN
```

Also, runs were made with and without this code to determine if the `ieee_is_nan()` call was expensive or not. The results are below with Throughput (days/day): 

| Resolution | Stock | IEEE |
| - | - | - |
| c48 | 569.414  | 562.533 |
| c90 | 216.222  | 219.643 |
| c180 | 126.599  | 128.268 |
| c360 | 112.156  | 113.598 |
| c720 | 44.752 | 45.567 |

As can be seen there is little cost to the calls at least on the order of discover variablility. Wall Time shows a similar trend. Per @tclune, the expected expensive bit (initializing the logger) is done in the called-once `ESINIT` code.

**NOTE**: In truth, this should probably be a logging *error* not warning, but the best way to propagate errors up the stack is through the `_ASSERT()` and `_VERIFY()` macros. This code doesn't support this as they do not return RC codes. But, since a NaN should crash the code right after this, in a way this does...

### DIFF
--- a/GEOS_Shared/GEOS_Utilities.F90
+++ b/GEOS_Shared/GEOS_Utilities.F90
@@ -153,6 +153,7 @@
       real, save :: TMINICE    =  ZEROC + TMINSTR
 
       class(Logger), pointer :: lgr
+      logical :: debugIsEnabled
 
   contains
 
@@ -507,8 +508,10 @@
     real    :: URAMP, DD, QQ, TI, DQ, PP
     integer :: IT
 
-    if (ieee_is_nan(TL)) call lgr%warning(' QSAT0: TL contains NaN')
-    if (ieee_is_nan(PL)) call lgr%warning(' QSAT0: PL contains NaN')
+    if (debugIsEnabled) then
+      if (ieee_is_nan(TL)) call lgr%warning(' QSAT0: TL contains NaN')
+      if (ieee_is_nan(PL)) call lgr%warning(' QSAT0: PL contains NaN')
+    end if
 
     if(present(RAMP)) then
        URAMP = -abs(RAMP)
@@ -526,11 +529,12 @@
        PP = PL*100.
     end if
 
-    if(URAMP==TMIX .OR. URAMP==0. .and. UTBL) then
+    if((URAMP==TMIX .OR. URAMP==0.) .and. UTBL) then
 
        if(FIRST) then
           FIRST = .false.
           call ESINIT
+          call LOGGER_INIT
        end if
 
        if    (TL<=TMINTBL) then
@@ -565,6 +569,11 @@
 
     else
 
+       if(FIRST) then
+          FIRST = .false.
+          call LOGGER_INIT
+       end if
+
        TI = TL - ZEROC
 
        if    (TI <= URAMP) then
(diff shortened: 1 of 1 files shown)


tclune:
I worry slightly that the IEEE numbers are actually a bit faster in this limited sampling (except C48, but who cares  there).     Presuming this is noise, Bill might care about an O(1%) impact that we cannot detect with out finer granularity.   

One way to get a more accurate timing would be to add  a new-fangled MAPL timer around the call.   Get the time spent.  Then make the IEEE call twice in the same location  and get the time spent.   The delta for the two runs should be very accurate compared to the noise above.

mathomp4:
> One way to get a more accurate timing would be to add a new-fangled MAPL timer around the call. Get the time spent. Then make the IEEE call twice in the same location and get the time spent. The delta for the two runs should be very accurate compared to the noise above.

@tclune I'll try that...though expect a ping on Teams when I get confused. I'll stare at the demos/tests in profiler and MAPL...

mathomp4:
Hmm. Yeah, @tclune might be right. From runs at C90 for one day. If we have just two `ieee_is_nan()` calls:
```
                                                                Inclusive        Exclusive
                                                             ================ ================
Name                                               #-cycles   T (sec)    %     T (sec)    %
                                                   --------- --------- ------ --------- ------
----------MOIST                                          384   113.890  25.21    82.218  18.20
------------IEEED3                                       576     0.082   0.02     0.082   0.02
------------DQSAT0                                  22091065    10.668   2.36     8.718   1.93
--------------IEEED0                                22091065     1.951   0.43     1.951   0.43
------------IEEE2                                        192     0.001   0.00     0.001   0.00
------------QSAT0                                   42497658    20.855   4.62    17.056   3.78
--------------IEEE0                                 42497658     3.799   0.84     3.799   0.84
------------IEEE1                                      43200     0.034   0.01     0.034   0.01
------------IEEE3                                        192     0.031   0.01     0.031   0.01
----------TURBULENCE                                     576     5.870   1.30     4.178   0.92
------------DQSAT0                                   3404369     1.692   0.37     1.380   0.31
--------------IEEED0                                 3404369     0.312   0.07     0.312   0.07
```
where `IEEE(D)N` profiles the `ieee_is_nan()` calls in (D)QSATN

mathomp4:
Also, I've found some uncovered code paths that would cause issues if anyone wanted to change how qsat behaves. Making more general.

mathomp4:
Okay. I think this is good now. I had to make a new `LOGGER_INIT` procedure because there were some code paths that weren't covered before. Now, I think the original way was 99.99999% fine, but if someone decided to not use the table lookup, I'm guessing bad things might happen.

sdrabenh (approved):

PR #134, opened 2020-10-22
Edits to make plots/configure portable

mathomp4:
One issue with running someone else's experiment is that if you do plots, you eventually need to set up links that are done with `plots/configure`. But someone else can't run `configure` in your build. 

This commit will do that. I had to do a few things like *hardcode* `GEOSUTIL` because...well...it has to be. We need the right `GEOSUTIL` so the links are in the right place!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added Docker authentication for CI
 
 ### Fixed
+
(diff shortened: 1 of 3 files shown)


mathomp4:
Note: I'm adding a blocker to this only so we can have @sdrabenh or @lltakacs make sure it works. I know @lltakacs was able to confirm a similar set of changes worked in CVS land, so I'm hoping these work in Git Land. 😄

sdrabenh (approved):

bena-nasa (approved):

PR #136, opened 2020-11-12
add dependency to turn on BUILD_SHARED_LIBS

weiyuan-jiang:


tclune (changes requested):
As per the other ticket - why is there a new dependency?

Also, please delete lines rather than commenting them.  (Lines 146 and 249)

weiyuan-jiang:
Again, this is not new dependency. We got away with it because we never use some subroutines

tclune (dismissed):

sdrabenh (dismissed):

mathomp4:
I am putting a blocker tag on this until I can consult with @weiyuan-jiang about the GNU/bufrlib issue.

mathomp4 (approved):

PR #143, opened 2020-12-09
Changes needed to allow use of aggressive flags with GEOS

mathomp4:
Activate by using `-DCMAKE_BUILD_TYPE=Aggressive`


mathomp4:
I tried this without the changes to ESMA_cmake and ecbuild and it was happy.

tclune (dismissed):
I think a character is required in column 6

mathomp4:
> I think a character is required in column 6

I think in [fixed-form](https://www.openmp.org/spec-html/5.1/openmpsu17.html#x32-310002.1.1) it is, but `mod_four.F90` is hopefully free-form, and so [the standard doesn't seem to require that](https://www.openmp.org/spec-html/5.1/openmpsu17.html#x32-310002.1.1).

tclune:
Oops.   I actually checked before I hit return.  Thought I had deleted the comment, but apparently not.  Sorry.

sdrabenh (approved):

bena-nasa (approved):

PR #146, opened 2020-12-14
Add partition option for regrid

mathomp4:
This PR adds a `-partition` option to `regrid.pl`.

Note that this PR is dependent on https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/392 as we need to pass the `-partition` down to `mk_restarts`


mathomp4:
I am marking this ready for review, but I've added a blocker to it as we should wait for https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/392 to be accepted first as the `-partition` option isn't fully supported until that is in.

sdrabenh (approved):

PR #153, opened 2021-01-20
Plot and stat updates from ltakacs on 2021-01-15

sdrabenh:


mathomp4:
> @mathomp4 and @bena-nasa could you please review the changes to stats.F90? Larry mentioned these came from Ricardo but it is unclear to me whether these are compatible with the current MAPL. Please advise.

Well, you do call `ESMF_Initialize()`in that program, so calling `ESMF_Finalize()` before/instead of `stop` does seem like a good idea. I'll let @bena-nasa be the final arbiter as he knows ESMF better than I.

bena-nasa (approved):

mathomp4 (approved):

Issue #99, opened 2020-06-09
Edge level arrays out-of-bounds in Lighting_mod


atrayano:
@christophkeller In GEOS, the indexing convention for edge variables is 0:LM. Lightning_mod.F90 references these 3 variables: PLE, cnvMfc, geoPotHeight (passed in from Moist as ZLE) from 1:LM+1 - around lines 916-928 in Lightning_mod.F90
This issue was exposed by requesting MAPL to always allocate all exports, which essentially enabled calling get_hemcoFlashrate - equivalently we could have asked for LFR_GCC in History, and compiling GEOS with debug flags. The compiler flagged cnvMfc, but all 3 vars have the same issue

mathomp4:
@christophkeller @atrayano  Has this been taken care of?

atrayano:
@mathomp4  not yet (thanks for reminding me)

christophkeller:
I just updated the lightning module and made a PR for the fix: https://github.com/GEOS-ESM/GMAO_Shared/pull/159
I'm also tagging @megandamon as the original creator of the lightning wrapper.

mmanyin:
@atrayano and @christophkeller  - Actually the original code was not in error, since all 3 of those fields are deferred shape arrays, and as such, have starting index 1.  The edge-array convention may have been violated, but the code was indexing correctly.  (BTW, the Lightning module development was handed off from Megan to me, and I have updates (working in Icarus) that should be incorporated into GIT, when time allows.)

PR #159, opened 2021-02-16
Bug fix for issue #99: Edge level arrays out-of-bounds in Lighting_mod

christophkeller:
This is a bug fix for the following issue: https://github.com/GEOS-ESM/GMAO_Shared/issues/99

Arrays ple and cnvMfc are now referenced using the indices LM:0 (instead of LM+1:1):

```
      pleHemco(:,:,1:KM) = ple(:,:,LM:0:-1)
      cnvMfcHemco(:,:,1:KM) = cnvMfc(:,:,LM:0:-1)
```

Similarly, the index was shifted by -1 for array geoPotHeight so that it now spans the valid range:

```
      do levCount=1,LM
         gridBoxHeightHemco(:,:,levCount) = geoPotHeight(:,:,levCount-1) - geoPotHeight(:,:,levCount)
      enddo
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+
+- Fix for out-of-bounds error in lightning module (#99)
+
 ### Changed
 ### Added
 ### Removed
--- a/GEOS_Shared/Lightning_mod.F90
+++ b/GEOS_Shared/Lightning_mod.F90
@@ -913,8 +913,8 @@ end subroutine getLightning
       ALLOCATE(ltop2d (1:IM, 1:JM),   STAT=RC)
 
       ! edge layer fields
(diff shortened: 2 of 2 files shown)


atrayano:
This fix is OK, I was thinking we can be more general, and use LBOUND/UBOUND to be absolutely sure. But for GEOS, this fix is totally fine. Thanks!

mathomp4:
@christophkeller Is this zero-diff or non-zero-diff?

atrayano:
@mathomp4 This is bugfix - if the lightning code was exercised, if would have crashed, while new PR fixes it. If it is not exercised, then of course it should be a zero-diff

mathomp4:
Okay. Just wondering how to label it. And I suppose I can figure out a Changelog message for this ("make code work").

christophkeller:
The bug fix prevents an out-of-bounds error, so maybe this could be the Changelog message ("prevent out-of-bounds error")

sdrabenh (approved):

PR #162, opened 2021-02-25
Update binarytile.x for river routing

bena-nasa:
This PR is contingent on this PR in MAPL:
https://github.com/GEOS-ESM/MAPL/pull/725

This PR is needed for eventual implementation of the river routing model. The river model will require modification to the tile files to include the number of pfafstetter catchments on the first line as the 2nd number. Right now binarytile.x was copying one number from the first line (number of tiles) to the binary. This will copy two if they exist. Note that this should not break anything as no one has ever previously cared about the other numbers that may be there which is the raster size used to create the file. Sarith said no utility uses this information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Changed
 ### Added
+
+- Extend binary tile to supprot future river route component development
+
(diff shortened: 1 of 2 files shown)


mathomp4:
You have a few `bmaa` writes in there. You might want to remove them!

bena-nasa:
Thanks! I removed the comments

sdrabenh (approved):

PR #163, opened 2021-02-28
Sync atmosphere-ocean interface with GEOS FP system: GEOS-5.27.1

sanAkel:
This update was implemented as `TSKINWTR patch` in GEOS FP system [GEOS-5.27.1](https://gmao.gsfc.nasa.gov/news/geos_system_news/2021/planned_GEOS_FP_upgrade_5_27_1.php)
Therefore it can be found (per @rtodling) in:
>CVS tag:  GEOSadas-5_27_1_p3

Please see `GEOS_OpenWaterGridComp.F90`. Of course, _this_ git based model version has it all re-organized, but **zero-diff** following https://github.com/GEOS-ESM/GEOSgcm/releases/tag/v10.16.1

**Description of the update**:
- Limit `TW, TWMTS` to `-1.8C` whenever `FR(N) <= fr_ice_thresh`. Note: the `default` value of  `fr_ice_thresh` is 0.0 and `FR` is positive semidefinite, i.e., it is **never** < 0.

This addition is non-zero-diff, and essentially syncs the atmosphere-ocean interface layer with what is being used by the [GMAO FP system GEOS-5.27.1](https://gmao.gsfc.nasa.gov/news/geos_system_news/2021/planned_GEOS_FP_upgrade_5_27_1.php), and is the only reason for this PR.

#### Notes:
This update has actually exposed 2 issues (1 already known):
1. [**known issue**] that the internal state variables: `TW, TWMTS` are _poor_ choices! which have already been replaced by a single _better_ one: `TMWTF`, used in the GMAO S2S v3; however, the FP system does not, reasons were [summarized in this issue](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/356).
2. [**now exposed**] Regardless of the choice of the internal state, in `GEOS_OpenWaterGridComp.F90` the fraction of water `FR` should be > 0, since its job is to operate over variables defined on water, again, where `FR water > 0`, needless to say, `FR water == 0` means its sea ice! So this issue brings to our attention the _need_ to treat variables where there is sea ice in  `GEOS_OpenWaterGridComp.F90`, which by design is the job of the sea ice child. Clearly either the parent: `GEOS_SaltWaterGridComp.F90` and/or its two children namely: `GEOS_SimpleSeaiceGridComp.F90, GEOS_OpenWaterGridComp.F90` are flawed! It is possible that this issue manifests in coupled simulations as well, where the sea ice child is: `GEOS_CICE4ColumnPhysGridComp.F90` instead of `GEOS_SimpleSeaiceGridComp.F90`.


CHANGELOG.md
```diff
@@ -33,6 +33,11 @@ be deleted in future releases of MAPL
 
 - Fix flags for `zonal.f` compilation
 
+
+### [version ??] - 2021-02-28
```
mathomp4, line 37: @sanAkel, This is why we have the "Unreleased" section at the top of the CHANGELOG. Please place all changes to this file there. We will no be issuing a version of GMAO_Shared between 1.3.6 and 1.3.7.

sanAkel: Thanks @mathomp4 got it, made the change. I will leave the version number for you to decide!

mathomp4: Thanks! Looks great!

sdrabenh (approved):

PR #164, opened 2021-03-01
Remove CMIP_1977_1982 directory

mathomp4:
Per @sanAkel, it appears the `CMIP_1977_1982/` directory in `GEOS_Util/pre/NSIDC-OSTIA_SST-ICE_blend/` is not needed. It was found in a sweep of Python files for the upcoming Python 3 Transition. And if @sanAkel says it's not needed, it's less code to carry around!


sanAkel (approved):
Thanks! @mathomp4 

The reason(s) why we don't need them anymore is because they originated from creation of `MERRA-2` SST and Sea Ice Concentration boundary condition preparation. 

1. As one can see, these scripts are **not** useful for near-real time. Of course, MERRA-2 is _now_ (03/2021) in near real time.
2. A new reanalysis would most certainly not use this processing chain, given what we already know about the _quality_ of the data. The reader is pointed to section `8.A` of https://gmao.gsfc.nasa.gov/pubs/docs/Bosilovich803.pdf

So, going forward, deletion is certainly carrying lighter weight.

tclune (approved):
Entropy is reduced!

sdrabenh (approved):

PR #165, opened 2021-03-16
Plot updates for stats from L.Takacs

sdrabenh:


mathomp4 (approved):
Approving since I trust @sdrabenh and @lltakacs on these things. :)

PR #170, opened 2021-04-08
Fixed CMake bugs for DAS build only libraries

bena-nasa:
When trying to build the DAS with the latest GMAO_Shared they do not sparse out arpack and the cmake command was failing. This fixes it.


tclune (approved):

bena-nasa:
I found more cmake problems, just using this PR instead of making a new one

tclune (dismissed):

mathomp4 (approved):

PR #175, opened 2021-04-16
Update main regrid.pl a la #174

mathomp4:
This is a `main` line update to `regrid.pl` to mirror that from @gmao-jstassi in #174. Per #174:

> Fixed the -wemin and -wemout options so that they will accept integer values; Also added Jason-NL BCS tag choice


rtodling:
This looks good to me. But there are no Reviewers on this!

sdrabenh (approved):

PR #197, opened 2021-06-21
Fixes some CMake Fortran_MODULE_DIRECTORY

mathomp4:
This is some CMake changes mainly to prevent `.mod` files from being in the `build/src` directory tree. 

Most of GEOS will put its `.mod` files in a `build/include` directory based on the library being made. So, for example, the `.mod` files for Moist are in `build/include/GEOSmoist_GridComp/`. And the GEOS CMake framework handles this 99% of the time. But in a few places, we find `.mod` files not yet in the `build/include` tree. This fixes those.

This is one of those "If it builds, it's zero-diff" PRs since the only changes are to the build and not to any source codes.


mathomp4:
Also, skipping the CHANGELOG as, well, this is so trivial it doesn't need to be noted.

tclune (approved):

PR #204, opened 2021-07-12
change the CN_VERSION and restart name

weiyuan-jiang:
1) This regrid.pl depends on some mk_restart programs in GEOSgcm_GridComp. This is not desirable and needs further discussion to re-organize the codes. 
2) @gmao-jkolassa,  I don't know perl and don't know how to use regrid.pl but I made changes anyway.  So I really need help to test the program with this branch https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/wjiang/jkolassa_cnclm45_mk_restart . I believe what we need for the command line is the CN_VERSION to be "clm40" or clm45"

### DIFF
--- a/GEOS_Util/post/regrid.pl
+++ b/GEOS_Util/post/regrid.pl
@@ -1579,9 +1579,25 @@ sub check_rst_files {
             else       { $notfound{$type} = 1 }
         }
         foreach $type (keys %notfound) {
-            next if $type eq "catchcn${cnlist[0]}_internal_rst";
             $SURFACE{$type} = 0;
             $swFLG = 1;
+            if ($type eq "catchcn${cnlist[0]}_internal_rst") { 
+               $SURFACE{$type} = 1;
+               if (scalar(@cnlist) eq 4) {
+                  $fname = rstname($cnlist[1], $type, "%s.%s.${ymd}_0000");
+                  my $ldas_rst_dir = "$cnlist[2]/$cnlist[1]/output/$cnlist[3]/rs/ens0000/Y${year}/M${month}/";
+                  $file = findinput($fname, $ldas_rst_dir);
+               }
+               elsif (scalar(@cnlist) eq 1) {
+                  $fname = rstname($expid, $type, "%s.%s.${ymd}_0000");
+                  $file  = findinput($fname);
+               }
+              if ($file) { $input_restarts{$type} = $file}
+              else  {
+                    die("File $file not found;");
+              }
+              next;
+            }
             if ($type eq "saltwater_internal_rst") {
                 $swFLG = 0 if $notfound{"openwater_internal_rst"};
                 $swFLG = 0 if $notfound{"seaicethermo_internal_rst"}
@@ -2906,7 +2922,8 @@ sub regrid_surface_rsts {
         $catch .= ".$tagID.$gridID" if $label;
 
         if ($mk_catchcn) {
-           $catchcnName = rstname($expid2, "catchcn${cnlist[0]}_internal_rst", $template2);
+           #$catchcnName = rstname($expid2, "catchcn${cnlist[0]}_internal_rst", $template2);
+           $catchcnName = rstname($expid2, "catchcn${cnlist[0]}_internal_rst", "%s.%s.${ymd}_${hr}z.nc4");
            $catchcnIN = "$InData_dir/$catchcnName";
            $catchcn = "$rstdir2/$catchcnName";
            $catchcn .= ".$tagID.$gridID" if $label;
@@ -3041,6 +3058,7 @@ sub rename_surface_rsts {
         }
         $newname = rstname($expid2, $type, "$outdir/$template2");
         $newname .= ".$tagID.$gridID" if $label;
+        if ($type eq "catchcn${cnlist[0]}_internal_rst" ){ $newname = rstname($expid2, $type, "$outdir/%s.%s.${ymd}_${hr}z.nc4")};
         move_("\n$newrst", $newname);
     }
 
--- a/GEOS_Util/post/regrid.pl
+++ b/GEOS_Util/post/regrid.pl
@@ -1584,12 +1584,12 @@ sub check_rst_files {
             if ($type eq "catchcn${cnlist[0]}_internal_rst") { 
                $SURFACE{$type} = 1;
                if (scalar(@cnlist) eq 4) {
-                  $fname = rstname($cnlist[1], $type, "%s.%s.${ymd}_0000");
+                  $fname = rstname($cnlist[1], $type, "%s.%s.${ymd}_${hr}00");
                   my $ldas_rst_dir = "$cnlist[2]/$cnlist[1]/output/$cnlist[3]/rs/ens0000/Y${year}/M${month}/";
                   $file = findinput($fname, $ldas_rst_dir);
                }
                elsif (scalar(@cnlist) eq 1) {
-                  $fname = rstname($expid, $type, "%s.%s.${ymd}_0000");
+                  $fname = rstname($expid, $type, "%s.%s.${ymd}_${hr}00");
                   $file  = findinput($fname);
                }
               if ($file) { $input_restarts{$type} = $file}
--- a/GEOS_Util/post/regrid.pl
+++ b/GEOS_Util/post/regrid.pl
@@ -1584,9 +1584,16 @@ sub check_rst_files {
             if ($type eq "catchcn${cnlist[0]}_internal_rst") { 
                $SURFACE{$type} = 1;
                if (scalar(@cnlist) eq 4) {
+                  # find GEOSldas file first
                   $fname = rstname($cnlist[1], $type, "%s.%s.${ymd}_${hr}00");
                   my $ldas_rst_dir = "$cnlist[2]/$cnlist[1]/output/$cnlist[3]/rs/ens0000/Y${year}/M${month}/";
                   $file = findinput($fname, $ldas_rst_dir);
+                  # find LDASsa file 
+                  unless ($file) {
+                     $fname = rstname($cnlist[1], $type, "%s.ens0000.%s.${ymd}_${hr}00z");
+                     $ldas_rst_dir = "$cnlist[2]/$cnlist[1]/output/$cnlist[3]/rs/ens0000/Y${year}/M${month}/";
+                     $file = findinput($fname, $ldas_rst_dir);
+                  }         
                }
                elsif (scalar(@cnlist) eq 1) {
                   $fname = rstname($expid, $type, "%s.%s.${ymd}_${hr}00");
--- a/GEOS_Util/post/regrid.pl
+++ b/GEOS_Util/post/regrid.pl
@@ -2824,7 +2824,7 @@ sub regrid_surface_rsts {
 
                 # use catch for catchcn, if needed. WJ note: use catch for cnclm40 only
                 #---------------------------------
-                if ($type eq "catchcnclm40_internal_rst") {
+                if ($type eq "catchcn${cnlist[0]}_internal_rst") {
                     $alt = "catch_internal_rst";
                     $src = $input_restarts{$alt};
                     if ($src) {
@@ -2851,6 +2851,12 @@ sub regrid_surface_rsts {
         symlinkinput($tile1, $InData_dir);
         symlinkinput($tile2, $OutData_dir);
     }
+
+    # link clsm directory to OutData
+    #-------------------------------
+    $clsm = dirname($tile2) ."/clsm";
+    $clsm = "$H2{bcsdir}/clsm" unless -d $clsm;
+    symlinkinput($clsm, $OutData_dir);
 
     # link rst directory to OutData
     #------------------------------
@@ -2915,12 +2921,6 @@ sub regrid_surface_rsts {
 
         move_("\n$catch", "$catchIN", $verbose)     if $mk_catch;
         move_("\n$catchcn", "$catchcnIN", $verbose) if $mk_catchcn;
-
-        # link clsm directory to OutData
-        #-------------------------------
-        $clsm = dirname($tile2) ."/clsm";
-        $clsm = "$H2{bcsdir}/clsm" unless -d $clsm;
-        symlinkinput($clsm, $OutData_dir);
 
         # catch and/or catchcn restart only during this pass
         #---------------------------------------------------
--- a/GEOS_Util/post/regrid.pl
+++ b/GEOS_Util/post/regrid.pl
@@ -918,15 +918,19 @@ sub check_inputs {
     #------------------
     @cnlist = split(/,/, $mk_catchcn) if $mk_catchcn;
     if ($mk_catchcn) {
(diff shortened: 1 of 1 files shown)


gmao-jkolassa:
Thanks @weiyuan-jiang . I'll try running regrid.pl using your modifications and will let you know if I encounter issues.

weiyuan-jiang:
@bena-nasa , would you please help test this program for the GCM's general use?

weiyuan-jiang:
@gmao-jkolassa , Ben has helped test other restarts, we only need to test catch, catchcn. To test catchcn, the CN_VERSION needs to be "clm40" or  "clm45"

gmao-jkolassa:
@weiyuan-jiang I tried to run regrid.pl by calling it directly. The command to do so can be found in /discover/nobackup/jkolassa/regrid_test/test2.CMD and the outputs are in /discover/nobackup/jkolassa/regrid_test/test_out/, which also contains the log file C90CS_f522_fp.20190614_21z.log

The problem seems to be that at some point regrid.pl is looking for the catchcnclm45_internal_rst file in the OutData directory and is unable to find it (The relevant line in regrid.pl is 2917). I _think_ the program that should move the file there is mk_Restarts. I also looked at the log file for mk_Restarts (/gpfsm/dnb31/jkolassa/regrid_test/test_out/mk_catch.log.1) and it seems to fail just at the point where it would process the catchcn restarts. I took a look at mk_Restarts (the relevant code seems to be on lines 285-307) and did not see anything obviously wrong and I thought that you had programmed the change in the name such that changes to mk_Restarts are not necessary, but maybe we missed something.

gmao-jkolassa:
I should correct my previous statement: mk_Restarts does not fail, but it only creates the catch_internal_rst in OutData and not the catchcnXXXX_internal_rst. I think that this may be because the relevant code segment has the following if-statement before processing the catchCN restarts 'if ($cnlen[0] == 4 && \$rescaleFLG) then', but rescaleFLG is 0, even though 'rescale=yes' in regrid.pl

weiyuan-jiang:
@gmao-jkolassa ,  There is an error in your command line test2.CMD. please remove the extra space between "," and "SMAP_EASEv2_M36". It should be just ",SMAP_EASEv2_M36" . I cannot access some of your directories and cannot not test it by myself. Please test it again and let me know what happens

gmao-jkolassa:
Thanks for catching that @weiyuan-jiang ! I did repeat the experiment after removing the space, but unfortunately I still get the same error message. I also gave you permissions to access /discover/nobackup/jkolassa/regrid_test/ . Please let me know if you need access to any other directories.

gmao-jkolassa:
Changed permissions.

As I was doing that, I realized that I actually have a typo in my test2.CMD. The year should be 2016, bot 2019. Let me try that again with the year corrected.

weiyuan-jiang:
ok, please let me know what happens after your changes.

gmao-jkolassa:
Hi @weiyuan-jiang ! I tested this again after correcting the date and I still get the same error message. I actually tested this two ways (all path and file names are relative to /discover/nobackup/jkolassa/regrid_test/):

(1) By providing the path and name of the catchCN restart file as a command line input to regrid.pl (command is in test5.CMD, inputs are in ./test_in2 and outputs are in ./test_out_2)
(2) By placing a restart file of the expected name in the input directory  (command is in test4.CMD, inputs are in ./test_in and outputs are in ./test_out).

Both experiments fail. Part of the problem seems to be that the script is looking for a binary catchCN restart file, whereas I only have netcdf restart files for CN.

weiyuan-jiang:
@gmao-jkolassa , please let me access the directory /discover/nobackup/jkolassa/GEOSldas45_M36_rst_ldas06_16/output/SMAP_EASEv2_M36/rs/ens0000/Y2017

gmao-jkolassa:
Sorry about that! Changed permissions.

weiyuan-jiang:
@gmao-jkolassa , Sarith made a lot of changes including this line:  https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/57857bea4ac165eaa1996e7fe74eeaac4d147a58/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_Restarts#L285
I believe it is a bug. The second condition ( rescaleFLG) should not be there.

gmao-jkolassa:
@weiyuan-jiang I agree. That is exactly the line that confused me too. As I mentioned in a comment above, 'rescaleFLG' is 0 for some reason (even though the main regrid.pl script has 'rescale' set to 'yes'), so the processing of the catchcnclm restarts does not get triggered. I can try to remove the rescaleFLG condition and see what happens.

gmao-jkolassa:
@weiyuan-jiang I changed the if-statement in question, but running regrid.pl still fails. The script fails at finding the catchcnclm45_internal_rst even if it is present in the input folder. I think the issue here is that it is looking for a binary file, but the catchcnlm45 restarts that I have are all netcdf files.

weiyuan-jiang:
@gmao-jkolassa . This didn't happen last time. In the codes, mk_catch and mk_catchcn often go together. Maybe the file type should be the same if regriding them together.

gmao-jkolassa:
@weiyuan-jiang This has happened in my tests before. I just did a quick test where I did provide a CatchCN4.5 binary restart file (and renamed it for the correct date and to match the catchcnclm45_internal_rst naming convention), but the run still fail in 'Catch Restart: step 2). The command is in test5.CMD, inputs are in ./test_in2 and outputs are in ./test_out_2.

I'll take another look tomorrow.

gmao-jkolassa:
@weiyuan-jiang I think I may have found the issue (or at least an issue). In regrid.pl on line 2827 I think the line 
` if ($type eq "catchcnclm40_internal_rst")`
should instead be 
` if ($type eq "catchcn${cnlist[0]}_internal_rst") `
I am testing this now and will let you know.

weiyuan-jiang:
When it regrids cnclm45, it is looking for cslm/catchment.def.   But I cannot find cslm directory through your command line input ( test4.CMD or test5.CMD). Maybe we are missing some BC parameters

gmao-jkolassa:
Well, in the command line input you are specifying the boundary conditions the input files used and which boundary conditions to use for the output files. In the command line inputs, those are specified through `-tagin` and `-tagout`. In our case, the input files used Icarus (ICA) and we want to use Icarus-NLv3 (INL) for the outputs. It is also possible to run the script interactively and select the BCs then. In any case, that should be how the script is told where to look for clsm/catchment.def.

weiyuan-jiang:
@gmao-jkolassa , I can tell from my error message that we didn't provide the BCS for the output.  I think Sarith never made it work. Would you please help me setup GEOSldas for restart option 2?  I would like to know two cases: one is GEOSldas restart the other is LDASsa restart.  Then I can integrate our mk_GEOSldasRestart into regrid.pl.

forrtl: No such file or directory
forrtl: severe (29): file not found, unit 10, file /gpfsm/dnb32/wjiang/tmp/regrid_test/test_out_5/23154321/OutData/clsm/catchment.def
Image              PC                Routine            Line        Source
mk_GEOSldasRestar  000000000049705B  Unknown               Unknown  Unknown
mk_GEOSldasRestar  00000000004B8110  Unknown               Unknown  Unknown
mk_GEOSldasRestar  000000000046CC03  mk_geosldasrestar         338  mk_GEOSldasRestarts.F90
mk_GEOSldasRestar  000000000042B07E  MAIN__                    221  mk_GEOSldasRestarts.F90

weiyuan-jiang:
I run your test5_cmd. /gpfsm/dnb32/wjiang/tmp/regrid_test/,   the log file /gpfsm/dnb32/wjiang/tmp/regrid_test/test_out_5/mk_catch.log.1

gmao-jkolassa:
@weiyuan-jiang I wanted to give you a quick progress update. I spent yesterday and today debugging the regridding and I think I finally got it to work. There were bugs in regrid.pl, mk_Restarts and mk_GEOSldasRestarts. If it is ok with you, I will commit the necessary changes to your branch feature/wjiang/jkolassa_cnclm45_mk_restart for GMAO_Shared and GEOSgcm_GridComp and you can review them to make sure that I didn't mess anything up. For documentation purposes and to have an overview, I am also including a summary of the changes below. There were four things that needed to be changed:

(1) To fix the error with regrid.pl not finding clsm/catchment.def it was necessary to establish a symbolic link from OutData to the bcs/clsm directory. The script actually had a code block to do that
```
$clsm = dirname($tile2) ."/clsm";
$clsm = "$H2{bcsdir}/clsm" unless -d $clsm;
symlinkinput($clsm, $OutData_dir);
```
 but it was included after catchment.def is needed. So I just moved that block up to line 2854.

(2) The script then ran into an issue finding the file OnData/OutTileFile even though that file existed. The issue seems to have been that OutTileFile was a symbolic link pointing to the actual *.til file at the output resolution. I made changes to mk_Restarts and mk_GEOSldasRestarts to make OutTileFile an actual copy of the *.til file rather than a symbolic link. Note that this slows down the regridding by the time it takes to copy the *.til file

(3) The bugfix that we addressed earlier that involved changing 
`if ($type eq "catchcnclm40_internal_rst")`
to 
`if ($type eq "catchcn${cnlist[0]}_internal_rst")`
in regrid.pl 

(4) Another bugfix we talked about earlier that involved changing line
` if ($cnlen[0] == 4 && \$rescaleFLG) then`
to  
`if ($cnlen[0] == 4) then`
in mk_Restarts.

Please let me know if anything is unclear. Otherwise I will go ahead and commit those changes to your branch.

weiyuan-jiang:
@gmao-jkolassa , that is a great progress. Please commit to my branch. After I verify it, then I can merge it to your branch, and finally to the main ( or develop) branch.

gmao-jkolassa:
@weiyuan-jiang I merged the committed the changes for GMAO_Shared to your branch. For the changes in GEOSgcm_GridComp, I had forgotten that we had already merged your branch into mine, so I committed those changes directly to feature/jkolassa_cnclm45.

weiyuan-jiang:
@gmao-jkolassa , I think there are still issues. In /gpfsm/dnb32/wjiang/tmp/regrid_test/mk_catch.log.2  I saw "catchcnIN: Undefined variable".  Indeed , the job script /gpfsm/dnb32/wjiang/tmp/regrid_test/mk_catch.j.2 does not define catchcnIN

gmao-jkolassa:
Hmm, ok, let me take another look. I didn't see that error when I ran a test on my end. Maybe I forgot to commit one of the changes I made.

gmao-jkolassa:
@weiyuan-jiang I had to implement another bugfix in mk_Restarts, which I pushed to GEOSgcm_GridComp branch feature/jkolassa_cnclm45. Basically the line to set catchcnIN before the call to Scale_CatchCN was missing. I managed to run the regridding without any more error messages now. Could you please test whether it also works for you?

I still have to verify that the regridded restart looks reasonably close to the original restart file. I'll do that tomorrow.

weiyuan-jiang:
@gmao-jkolassa . There is no error message now. But I still have question why we need two catchcn restart files in two locations. One is with all the other restart file , the other is in the output of GEOSldas or LDASsa output.

gmao-jkolassa:
@weiyuan-jiang  As far as I can tell, regrid.pl checks for the presence of a binary catchcn restart file, but the only uses the GEOSldas or LDASsa generated netcdf file to actually generate the regridded restart. I will have to double-check this, but if it is true, then maybe we can remove the check for binary file. But in the end, Sarith wrote these scripts, so I can only guess that this is how he intended it to work.

weiyuan-jiang:
@gmao-jkolassa . Thanks. I am not familiar with his program either. But I have some time now and will take care of this. Wait for my changes...

weiyuan-jiang:
@gmao-jkolassa . This is help message in regrid.pl from Sarith. How about adding another option:  -catchcn CN_VERSION ? it would search for nc4 file under the restart directories together with other restart file. That will be easier for general GCM users. But I am not sure if it is doable. 

  -catchcn           offers 2 options:
                      = 0, cold start for CLM40 using an archived restart file. For e.g. 
                        -catchcn 0, - Notice the trailing comma
                      = CN_VERSION, RESTART_ID, RESTART_PATH, RESTART_DOMAIN 
              to start from a GEOSldas restart file (Note, keywords are comma-separated).
                        where CN_VERSION is the 2-digit CLM version (40, 45, etc.) while other keywords 
                        are same as those in GEOSldas exeinp file. For e.g. 
                        -catchcn 45,GEOSldas45_M36_rst_ldas06_16,/discover/nobackup/elee15/GEOSldas_4_5/sims/,SMAP_EASEv2_M36
                        IMPORTANT GEOSldas restart file at 0z on the AGCM restart date must be available in GEOSldas directory.  
                      Not valid for tags prior to Heracles; Note: This option will add
                      10-20 minutes to the regrid process.

gmao-jkolassa:
@weiyuan-jiang Yes, I agree, it's a good idea to make this easier to use for GCM folks. When you say 

> it would search for nc4 file under the restart directories together with other restart file

do you mean in the same directory where all the other restart files are located? Because I think in most cases that does not necessarily have a netcdf restart for CatchCN. Or do you mean that if the user specifies that option we would essentially hardcode the script to get a restart file from the directories we have been using, e.g. GEOSldas45_M36_rst_ldas06_16,/discover/nobackup/elee15/GEOSldas_4_5/sims/ for CatchCN-CLM45?

The important part with CatchCN is to use a restart file from one of the simulations that has been spun-up over several centuries, since otherwise the results don't make much sense. So, if the user cannot specify a restart file like that, then I think we should internally point to one. Is that what you had in mind?

From a coding perspective, it should not be hard to implement, but one issue might be that we have to put CatchCN restart files in a location that is not likely to change (instead of someone's nobackup directory as is the case now).

weiyuan-jiang:
@gmao-jkolassa With the new changes, users don' have to copy a catchcn restart file after specifying the location. But there are still two issues 1) user needs to create a link YYYYMMDD_0000 no matter what start hour is  2) cold restart option for catchcn is still not working. We can keep improving regrid.pl one by one.  Which issue do you think is more important?

gmao-jkolassa:
Thanks @weiyuan-jiang ! I think it is probably more important to focus on issue (1). Cold-start is really not recommended, so while it is important to make that option work also, it really should not be used and I don't think anyone would run this in the near future, so there is less urgency.

So let's do issues (1) and (2) in that order. Do you want me to take a look at (1) or are you going to do it?

weiyuan-jiang:
@gmao-jkolassa , I will take care of it. There are many places that need to be chaned.

weiyuan-jiang:
@gmao-jkolassa , This PR together with https://github.com/GEOS-ESM/GEOSldas/pull/436  and https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/449  will make all  branches of "feature/jkolassa_cnclm45" work together.  1) no need to create catchcn restart file under restart_in directory. 2) you may need a right restart name in GEOSldas's output structure. Please double test my PRs. Thanks.

gmao-jkolassa:
Thanks @weiyuan-jiang ! I will run some tests with this tomorrow and will report back.

PR #208, opened 2021-07-29
Add pyrob_CF script

mathomp4:
I was informed by @keknowland that there is a `pyrob_CF` script that the GEOS-CF folks use. This adds it here so we capture it on GitHub


tclune (approved):
cmake changes ok

sdrabenh (approved):

PR #224, opened 2021-10-14
Add support for GEOS-IT in pyrob

mathomp4:
This PR adds support to `pyrob` to support the proposed/new GEOS-IT files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Updated `pyrob` to work with GEOS-IT files
+
 ### Removed
 
 ## [1.4.10] - 2021-10-08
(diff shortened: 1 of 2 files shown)


rtodling (approved):
Seem good to me!

mathomp4:
@rtodling Thanks. I'm waiting to hear back from Rob (not on GitHub as far as I can see 😄) as to what the higher ups want for the format. What is here works, but they might want different sorting? Not sure. I sent Rob two different GEOS-IT Word docs to evaluate. 

Though if there is a GMAO_Shared to be released soon, perhaps it's better to just get this in and then fix up sorting.

rtodling:
Before we put out a final 5.29.3 tag (beyond the rt-... tag) we should work this in.

mathomp4:
@rtodling Okay. I'll speak with @sdrabenh about getting it in. It can't exactly hurt anything!

sdrabenh (approved):

PR #238, opened 2022-01-18
adding python version of regrid.pl

weiyuan-jiang:
Add python version of regrid wrapper.  Hopefully it will replace regrid.pl . More testing is needed.

Requires related PR: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/571


tclune (commented):
cmake changes ok

mathomp4:
@weiyuan-jiang I might want to work on you with this to make it more "in-line" with the Perl version. I made some updates recently in there and it would be nice to make this roughly the same.

sanAkel:
@weiyuan-jiang, @mathomp4 when we _meet_ on Monday, would you please walk me through _this_ so I can add in an option to let user pick path to his/her boundary conditions, tile file etc?

1. I would like to add that option, test for my use case
2. If you are okay, push it to _this_ branch

Thanks!

weiyuan-jiang:
> @weiyuan-jiang, @mathomp4 when we _meet_ on Monday, would you please walk me through _this_ so I can add in an option to let user pick path to his/her boundary conditions, tile file etc?
> 
>     1. I would like to add that option, test for my use case
> 
>     2. If you are okay, push it to _this_ branch
> 
> 
> Thanks!

@sanAkel , Yes, I can help add those options to regrid.py. We would like your suggestion to improve this tool

gmao-rreichle:
Changing to "draft" PR for now.  FYI, the land group, with Weiyuan's support, will attempt to clean up the land portion of the [mk_restarts package in the GCM GC repo](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts), which includes loads of duplicate code and is in dire need of restructuring.  The mk_restarts effort and the present PR need to be synchronized.  This will likely take a while longer.

sanAkel:
> Changing to "draft" PR for now. FYI, the land group, with Weiyuan's support, will attempt to clean up the land portion of the [mk_restarts package in the GCM GC repo](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts), which includes loads of duplicate code and is in dire need of restructuring. The mk_restarts effort and the present PR need to be synchronized. This will likely take a while longer.

@gmao-rreichle may be we can proceed with whatever is in current format and attempt:

>will attempt to clean up the land portion of the [mk_restarts package in the GCM GC repo](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts), which includes loads of duplicate code and is in dire need of restructuring.

The **reason** for suggesting :arrow_up: is to get the code PR "merged" via @sdrabenh or I am afraid, this will take longer ... meaning, @GEOS-ESM/land-team's clean up of `mk_restarts`, then consideration by @sdrabenh. This is just a suggestion!

gmao-rreichle:
> > Changing to "draft" PR for now. FYI, the land group, with Weiyuan's support, will attempt to clean up the land portion of the [mk_restarts package in the GCM GC repo](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts), which includes loads of duplicate code and is in dire need of restructuring. The mk_restarts effort and the present PR need to be synchronized. This will likely take a while longer.
> 
> @gmao-rreichle may be we can proceed with whatever is in current format and attempt:
> 
> > will attempt to clean up the land portion of the [mk_restarts package in the GCM GC repo](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts), which includes loads of duplicate code and is in dire need of restructuring.
> 
> The **reason** for suggesting ⬆️ is to get the code in this present PR "merged" via @sdrabenh or I am afraid, this will take longer ... meaning, @GEOS-ESM/land-team's clean up of `mk_restarts`, then consideration by @sdrabenh. This is just a suggestion!

@sanAkel: I'm a bit unclear what exactly you're suggesting here.  If you're suggesting that we merge the new python version of `regrid`, my concern would be that it's going to be very difficult to test all possible use cases, of which there are a lot for land restarts.  Having the new  python version tied into the land-related `mk_restarts` cleanup would take care of this issue.  And yes, this is going to take time, unfortunately.  At the moment, I wouldn't quite know how to approve this PR for the land team (whose approval isn't actually required...).  We can of course discuss how progress on this PR relates to development needs by the ocean and/or landice teams.  It's not my decision, of course.  Just my 2c.

sanAkel:
> very difficult to test all possible use cases, of which there are a lot for land restarts.

I was not aware of such test cases! Since they exist, I support _holding_ this PR until ALL possible tests have been made.

While I say this, I would suggest that:
1. The modifications/additions within _this_ PR be tested, so as to make sure **ALL** tests are passing - incl. land restarts.
2. Once they pass, any _clean up_ can happen, **later**.

To be more clear:
- If tests in (above) 1 show that this code in this PR is failing, obviously, this PR remains in draft form!
- IMO I was hoping for an alternative to the `regrid.pl` through this python based handler. It seems to be working for _stock_ test case(s); I had no idea about all land restarts!
- Holding this work for 💯 % perfection can be avoided!

I see no required approver list (names) or number of approvals: **1** seems enough!! So all the :arrow_up: is just a request! 😄

gmao-rreichle:
>> very difficult to test all possible use cases, of which there are a lot for land restarts.

> I was not aware of such test cases! Since they exist,...

The problem is, they do **not** (yet) exist.  If they existed, I'd be happy to test and merge.  Right now, I'm not even sure we know use cases should work.  I certainly don't know.

GEOS_Util/post/regrider_surf.py
```diff
@@ -0,0 +1,337 @@
...
+
+     self.saltwater = None
+     self.openwater = None
+     self.seaice  = None
+     self.landice = None
+     self.lake  = None
+     self.route = None
+     self.catch = None
+     self.catchcnclm40 = None
+     self.catchcnclm45 = None
```
sanAkel, line 27: @weiyuan-jiang would you please split this up into following chunks:
- saltwater, openwater, seaice
- landice, lake
- route
- catch, catchcnclmxx; xx = 40, 45, ...

Of course, follow above suggestion through the rest of your script. I imagine addition of another script/module which would take input(s) from one of the above _chunks_ but does the **same** thing and writes output for that _chunk_. Makes sense?

sanAkel (changes requested):

sanAkel:
@gmao-rreichle I think I can see multiple issues. Please see above "request" that should be simple step to disengage things ... and honestly think its necessary!

GEOS_Util/post/regrider_surf.py
weiyuan-jiang: @sanAkel , I think it makes sense to break those up since they are using different executables .

sanAkel:
@weiyuan-jiang checking here ... any ETA when this can be un-drafted?

weiyuan-jiang:
> 

I will speed it up

sanAkel:
> > 
> 
> I will speed it up

🙏

GEOS_Util/post/regrid_catchment.py
```diff
@@ -0,0 +1,172 @@
...
+import shlex
+
+class catchment(object):
+  def __init__(self, params_file):
+     stream = open(params_file, 'r')
+     self.config = yaml.full_load(stream)
+     stream.close()
+
+  def regrid(self):
+     print("\nRegridding land, landice, saltwater.....\n")
```
gmao-rreichle, line 17: Shouldn't this be `print("\nRegridding land (catch)...\n")`
This script only deals with Catch and not landice or saltwater, right?  Would it work for CatchCN[xx]?

GEOS_Util/post/regrid_lakelandice.py
```diff
@@ -0,0 +1,143 @@
...
+import shlex
+
+class lakelandice(object):
+  def __init__(self, params_file):
+     stream = open(params_file, 'r')
+     self.config = yaml.full_load(stream)
+     stream.close()
+
+  def regrid(self):
+     print("\nRegridding land, landice, saltwater.....\n")
```
gmao-rreichle, line 17: `print("\nRegridding lake and landice...\n")`    ??

GEOS_Util/post/regrid_lakelandice.py
```diff
@@ -0,0 +1,143 @@
...
+           if (f.find(rst_name) != -1):
+              return f
+        return None  
+     
+     saltwater = find_rst(restarts_in,'saltwater')
+     openwater = find_rst(restarts_in,'openwater')
+     seaice  = find_rst(restarts_in,'seaice') 
+     landice = find_rst(restarts_in,'landice')
+     lake  = find_rst(restarts_in,'lake')
+     route = find_rst(restarts_in,'route')
```
gmao-rreichle, line 63: Why do we have catch, saltwater, etc when the script name is "regrid_lakelandice.py"?

tclune:
@weiyuan-jiang There appear to be a few unresolved issues in this PR that are on your plate.  Can you please summarize the status of each and provide an estimate for when you might be able to complete them?  Thanks.

GEOS_Util/post/regrider_surf.py
weiyuan-jiang: broke up excep Catchment which will be using the same executable

weiyuan-jiang:
> 

Tom, I tested one or two cases and it seems working. What I need is more tests. Without testing and benchmark, I don't have much confidence...

gmao-rreichle:
> There appear to be a few unresolved issues in this PR that are on your plate. Can you please summarize the status of each and provide an estimate for when you might be able to complete them? Thanks.

@tclune: I am working with @weiyuan-jiang on this and a [related PR in the GEOSgcm_GridComp](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/571).  The necessary changes are quite complex, owing to the current, messy state of the mk_restarts package.  Alas, I don't expect this to get done quickly, given the complexity and other, competing priorities.

GEOS_Util/post/regrid_py/regrid.py
```diff
@@ -43,6 +42,7 @@ def main(argv):
     params.convert_to_yaml()
     config_yaml = 'regrid_params.yaml'
 
+  sys.exit()
```
sanAkel, line 45: its good to set an:
- `error` message, e.g., `sys.exit('XYZ failed! Check ABC!')
- `successful` return, e.g., `sys.exit(0)`

sanAkel (changes requested):

GEOS_Util/post/regrid_py/regrid_catchment.py
```diff
@@ -18,67 +19,62 @@ def __init__(self, params_file):
...
-        return None  
-     catch = find_rst(restarts_in,'catch_')
-     if (not catch) :
-        print(" No catchment restart")
-        return
-
-     catch_f = os.path.basename(catch)
-     dest    = InData_dir+'/'+ catch_f
+     if not os.path.exists(InData_dir) :
+       print ("mkdir " + InData_dir)
```
sanAkel, line 64: `mkdir -p` would be better

weiyuan-jiang: sounds good. Added parameter in os.mkdirs to avoid inquiry

GEOS_Util/post/regrid_py/regrid.py
weiyuan-jiang: good

sanAkel:
> @sdrabenh Hi, Scott, do you have bench cases of regrid.pl ? I would like to test our regrid.py . @gmao-rreichle @biljanaorescanin

@rtodling and or @gmao-jstassi  may have some ... as well.

gmao-rreichle:
After talking with @tclune, I think we should proceed as follows:
1) Think of regrid.py as a new tool that should be released ASAP.  Besides the new script being in python, it also interfaces with new/revamped Fortran programs in the GEOSgcm_GridComp  Surface/Utils/mk_restarts directory.  (Hence the slower-than-hoped-for progress on this PR.)  
2) The existing regrid.pl should **not** be touched, nor should there be changes in any of the associated old Fortran programs in the GEOSgcm_GridComp  Surface/Utils/mk_restarts directory.  This ensures full compatibility with all existing setup and regridding scripts.  (The ultimate goal is to phase out the old regrid.pl and its underlying old Fortran programs.)
3) As of right now, regrid.py was successfully tested by @weiyuan-jiang for re-gridding (re-tiling) between a couple of cube-sphere grid resolutions and between a couple of EASE grid resolutions.  @biljanaorescanin still needs to test regrid.py when the bcs version changes (e.g., from Icarus to Icarus-NLv3, as in MERRA-2 to FP 5.25+) without and with also changing the resolution at the same time.  This will test parts of the code (Catchment prognostics rescaling) that have not yet been tested.  With any luck, this can be done within a week or so.
4) Once the land group is satisfied that regrid.py works for the above key use cases, we will ask @sdrabenh to do his testing of the PR (mostly verifying that regrid.pl indeed remains intact), merge the PR, and create a new tag.  
5) Next, we can solicit input from across GMAO on re-gridding (re-tiling) use cases, which can then be transitioned to regrid.py with the support of the SI team.
6) Eventually, we can remove regrid.pl and its old Fortran programs from the develop/main branch.
 
Feel free to chime in and correct the above as needed.

cc: @tclune @sdrabenh @sanAkel @mathomp4 @biljanaorescanin @weiyuan-jiang @rtodling @gmao-jstassi

biljanaorescanin:
Thanks @weiyuan-jiang , I will try again and let you know how it goes.

weiyuan-jiang:
The program search restart files under 

rst_dir/upperair
rst_dir/surface
rst_dir/analysis

It may not be convenient. Should we just leave all restart files under  rst_dir/  without subdirectories?

biljanaorescanin:
> The program search restart files under
> 
> rst_dir/upperair rst_dir/surface rst_dir/analysis
> 
> It may not be convenient. Should we just leave all restart files under rst_dir/ without subdirectories?

Either that or to list what each subdirectory has to have from the restart files. I doubt many users will know what to copy where if we don't list files.

sdrabenh:
@weiyuan-jiang I tried running this new script interactively and encountered the error below after the questionary section:

```
input tile file: /discover/nobackup/ltakacs/bcs/Icarus-NLv3/Icarus-NLv3_MERRA-2//CF0180x6C_DE1440xPE0720/CF0180x6C_DE1440xPE0720-Pfafstetter.til

output tile file: /discover/nobackup/ltakacs/bcs/Icarus-NLv3/Icarus-NLv3_Reynolds//CF0024x6C_DE0360xPE0180/CF0024x6C_DE0360xPE0180-Pfafstetter.til
Traceback (most recent call last):
  File "/discover/nobackup/projects/gmao/g6dev/sdrabenh/Models/GIT_PRs/PRs_20220517/install/bin/regrid.py", line 62, in <module>
    main(sys.argv[1:])
  File "/discover/nobackup/projects/gmao/g6dev/sdrabenh/Models/GIT_PRs/PRs_20220517/install/bin/regrid.py", line 41, in main
    params = regrid_params(config)
  File "/gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_20220517/install/bin/regrid_params.py", line 41, in __init__
    with  open('regrid_params.tpl', 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'regrid_params.tpl'
```
I confirm this path/file does exist, but the script cannot locate it internally. Is there something special that needs to be loaded first?

sdrabenh:
@weiyuan-jiang @tclune @gmao-rreichle @sanAkel @biljanaorescanin I have a few comments about the general flow of this new python script compared to the old pearl script after trying it. My first observation is that all the input questions appear to be grouped together, followed by all the output questions. For example, in the old pearl script:
```
...
INPUT atmosphere grid:
Enter OUTPUT atmosphere grid:
INPUT atmosphere levels:
Enter OUTPUT atmosphere levels:
...
```
This isn't necessarily a bad thing, but quite different than what most users are accustomed to. I'm just curious what motivated that change?

Secondly, in the modern age of netcdf and python, why can't we automatically read the restart files to know exactly what the input resolutions, data, etc. are? Even the old pearl script filled in the atmosphere grid resolution, but this new script does not. There are additional questions such as `Enter input restarts expid:` - is this necessary? Also, the `Enter new restarts expid:` is not pre-filled as the pearl script did. It seems to be an unnecessary burden to the user to fill in all these prompts. I fully admit I have not been part of the development discussion, so there could be a good reason for this. However, I would love to see less questions and more automatic detection/pre-filling for the interactive side of this script.

gmao-rreichle:
Good points, @sdrabenh, and thanks for test driving.  Re. the auto-detection of the grid in input restarts: This may not work for the tile-based restart files, not sure how much info we have in, say, Catchment restarts.  Ideally, of course, we don't have to ask.
Also, based on the snippet posted by @sdrabenh above, it looks like the user is not provided a sample of possible grid specs.  How would we tell the script that the (input or output) rst file is/should be in the EASEv2_M09 tile space?  "EASEv2_M09"?  "M09"?  Some other string?  For make_bcs, we offer a list of choices in the questionnaire.

weiyuan-jiang:
> 

Thanks for the comments. Will address your concern

weiyuan-jiang:
> why can't we automatically read the restart files to know exactly what the input resolutions,

My original thought was that the users don't need "fvcore_internal_rst" to do regridding of other restart files. But  we can make it more specific to gcm restarts if users go through questionnaire.  We have a verification though if the users input wrong input grid.

weiyuan-jiang:
> 
I have fixed this. Now the program should be able to run from different locations as long as regrid_params.py and regrid_params.tpl stay in the same location. I have fixed a few other bugs too.

sdrabenh:
> > why can't we automatically read the restart files to know exactly what the input resolutions,
> 
> My original thought was that the users don't need "fvcore_internal_rst" to do regridding of other restart files. But we can make it more specific to gcm restarts if users go through questionnaire. We have a verification though if the users input wrong input grid.

@weiyuan-jiang I'm not sure I understand your original thoughts, but I often use regrid.pl for changing vertical/horizontal resolution for the atmosphere/ocean from various sources for my gcm experiments. So I think in terms of the gcm. I agree that auto-detection may not be feasible for every question, tile files, or even for binary restarts (shudder). However, I'm a big proponent of using as much automation as possible rather than a straight-up list of blank questions. I have seen many users stumble with regridding because they don't know their source files. At the very minimum, if a user provides an input directory containing a single set of modern Jason/Icarus restarts, python commands could rather easily extract the EXPID, date, time, levels, atmospheric grid, etc. I would suspect there might be some key variables identifying the catchment or ocean types too - but admittedly that isn't my wheelhouse. For me, the ideal would be to have the script interrogate all the input files and list everything that could be identified for a user to confirm with a single "y", or alternatively go back and change an option if they knew otherwise. I suspect that might be difficult to implement at this point.

sdrabenh:
> > 
> 
> I have fixed this. Now the program should be able to run from different locations as long as regrid_params.py and regrid_params.tpl stay in the same location. I have fixed a few other bugs too.

Thanks! That is great. I will give it a try

weiyuan-jiang:
> 

@sdrabenh , That is a good point. I will let the program detect the information as much as possible. For example, it only asks for resolution if it could not find fvcore restart. It won't be hard to make such changes.

weiyuan-jiang:
I tested against mk_GEOSldasRestarts.F90 and it is zero-diff. But it is none zero-diff against mk_CatchRestart.F90 which is used by regrid.pl.  The non-zero diff variables are TSURF, TC(:,3), QC(:,3), CM, CH, FR, CQ . Those variable are re-set in mk_GEOSldasRestarts.F90 when regridding.  Sarith once told me mk_GEOSldasRestarts.F90 is more advanced comparing with mk_CatchRestarts.F90.  How do you think of those variables? I can make it zero-diff easily if you like.  @gmao-rreichle

gmao-rreichle:
> I tested against mk_GEOSldasRestarts.F90 and it is zero-diff. But it is none zero-diff against mk_CatchRestart.F90 which is used by regrid.pl. The non-zero diff variables are TSURF, TC(:,3), QC(:,3), CM, CH, FR, CQ . Those variable are re-set in mk_GEOSldasRestarts.F90 when regridding. Sarith once told me mk_GEOSldasRestarts.F90 is more advanced comparing with mk_CatchRestarts.F90. How do you think of those variables? I can make it zero-diff easily if you like. @gmao-rreichle

@weiyuan-jiang : Here are a couple of thoughts:
- The variables in question have little memory and spin up quickly, so small differences should be of little practical concern.
- TSURF should not be needed in the catch restart file because it is not a Catchment model prognostic variable.  I'm fairly sure that GEOSldas does not use it.  It is possible that the GCM or other apps rely on TSURF being in the catch restart file. 
- I don't know how TSURF, TC, QC, ... are in modified in mk_GEOSldasRestarts.F90 but not in mk_CatchRestart.F90s.  Can you point us to the exact place in the code where the differences originate?
- Assuming the modified TSURF, TC, QC, ... from mk_GEOSldasRestarts.F90 are more appropriate, then we would have to discuss with @sdrabenh and others whether it's preferable to include these changes in the regrid.pl perl script or whether it is preferable to maintain 0-diff for regrid.pl.  (We should definitely use the more appropriate treatment with the new regrid.py python script.)
- I remember @biljanaorescanin saying that some temperatures end up with 0 values when you get non-zero-diff with regrid.pl.  If this is indeed the case, then it would need to be fixed.  Earth system temperatures cannot be 0.  Presumably, this would be an issue with no-data-value handling.

biljanaorescanin:
> I remember @biljanaorescanin saying that some temperatures end up with 0 values when you get non-zero-diff with regrid.pl. If this is indeed the case, then it would need to be fixed. Earth system temperatures cannot be 0. Presumably, this would be an issue with no-data-value handling.

from output catch restart using .py code : 
`    -1 :       Date     Time   Level Gridsize    Miss :     Minimum        Mean     Maximum : Parameter name`
`    46 : 2022-05-16 21:00:00       0  4752080       0 :      0.0000      216.99      327.52 : TC `

from output catch restart  using .pl code:
 `    46 : 2022-05-16 21:00:00       0  4752080       0 :      238.61      285.21      327.52 : TC `

units for TC are kelvins so min zero with .py code is the issue.

weiyuan-jiang:
In mk_CatchRestarts.F90, every variable is re-tiled. But in mk_GEOSldasRestats.F90,  Some variables are reset:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3e52659a4baa55aea5b0c937afcbfe90f0fafd92/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_GEOSldasRestarts.F90#L2985-L3002

TC and QC only go to the first 3 : ( maybe it is a bug by Sarith? I am not sure)

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3e52659a4baa55aea5b0c937afcbfe90f0fafd92/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_GEOSldasRestarts.F90#L2835-L2869

TSURF is set to zero in catchcn ( I can correct it in catch model)

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3e52659a4baa55aea5b0c937afcbfe90f0fafd92/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_GEOSldasRestarts.F90#L1063

gmao-rreichle:
> In mk_CatchRestarts.F90, every variable is re-tiled. But in mk_GEOSldasRestats.F90, Some variables are reset:

My suspicion is that mk_GEOSldasRestats.F90 makes up WW, FR, CH, CM, CQ so that these variables are present in the restart in case the restart is used for the GCM, which needs these variables.  (GEOSldas does not need them.)  Before GEOSldas, LDASsa created catch restart files that did not have these variables, so if someone wanted to use an offline (LDASsa) restart for the GCM, they had to add these variables into the restart.  The made-up values are probably there for this reason. It's odd (but entirely surprising) that these variables are made up even if present in the input catch restart.  

Going forward, these variables should not be made up.  GEOSldas can create restarts that include these variables.  So regridding should use these variables from the input restart.  If they are missing from the input restart, they should be skipped (set to no-data, or just not included in the catch restart).   For the regrid perl script, this (probably) means that it should use mk_CatchRestart.F90 to maintain 0-diff.  For the regrid python script, the corresponding new F90 program in GEOS Catch GC should also not make up WW, FR, CH, CM, CQ. 

> TC and QC only go to the first 3 : ( maybe it is a bug by Sarith? I am not sure)

TC(4,:) and QC(4,:) are the temp and spec humidity over snow, which are diagnosed from the snow water equivalent, depth, and heat content.  In this sense, they are like TSURF and not model prognostic variables.  GEOSldas does not need TC(4,:) and QC(4,:) in the catch restart.  I assume the same is true for the GCM, but I can't say for sure. 

> TSURF is set to zero in catchcn ( I can correct it in catch model)

Not sure why this block sets variables to 0 rather than no-data.  It looks like these 0s are immediately written into the restart file.  I might be missing something here.  

cc: @rdkoster @sanAkel @sdrabenh @biljanaorescanin

weiyuan-jiang:
I made changes to https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/rreichle/mk_restarts_cleanup  . Let's satisfiy GCM's zero-diff first. @biljanaorescanin It should be zero-diff now even for those "not important" variables.

sdrabenh:
@weiyuan-jiang I am still getting runtime failures regridding. Please see an example interactive case in my directory. The complete logfile with inputs is located here: `/discover/nobackup/projects/gmao/g6dev/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python/output.log`
The complete log for the pearl script doing the exact same thing is here:
`/discover/nobackup/projects/gmao/g6dev/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_pearl/output.log`
The pearl script works but the python script does not. Also, the python script should recognize whether the user is on an interactive node a determine whether to submit a batch job accordingly. The pearl script is able to do this.

@tclune I do not think this script is ready yet

weiyuan-jiang:
> 

I know the problem here. 1) need to convert the relative path of restart dir to absolute path.  2) I can add the detection of interactive mode.

gmao-rreichle:
> @gmao-rreichle It's been more than the 1-2 weeks for testing since we last talked. Can you please summarize again here, what the remaining obstacles are and a reasonable timeline for acceptance for the initial release.

@tclune : The land group has been testing this PR since the two of us talked.  Unfortunately, testing turned up a few issues that need fixing.  The latest delay is because different results were produced depending on which of the underlying (existing) Fortran programs was connected to regrid, which pointed to a flaw in the (new) regrid python version.  The discrepancy was inherited from the previous author of the tools.  @weiyuan-jiang cannot be expected to understand the science issues, and he originally ended up with the wrong choice for regrid.py.  The good news is that the process is working, albeit at a snail's pace.  This is unfortunate, but the necessity to resolve this PR "by committee" just takes more time than we want it to take.  

The alternative is to merge code that we have not fully verified, which I am reluctant to do.  I am sorry, but not surprised, that this PR is taking a lot longer to resolve than anticipated.  I had communicated this expectation from the start.  Unfortunately, this does not bode well for resolving the required additional fixes that are needed in make_bcs. 

@rdkoster @biljanaorescanin @sanAkel

tclune:
OK - we have some serious process problems on the development of this tool.      I should have pushed harder for a minimum test suite when we (SI Team) decided to take this on.   In the mean time we have a PR  that now encompasses multiple bug reports and feature requests.     In a more perfect world, each of those should be their own issue, and we can then resolve them as we go.

So 1st step - I'm going to open a new issue to define a _minimal_ test suite, and notify the major players here.    Possibly this is something @gmao-rreichle et al already have, but I'm guessing that their suite is larger than what I envision.    Once the tool can pass this initial test suite, the PR should be merged.   We can then mine this ticket for new issues/features and open corresponding tickets.   (Happy to have others prioritize among them.)

sanAkel:
> OK - we have some serious process problems on the development of this tool. I should have pushed harder for a minimum test suite when we (SI Team) decided to take this on. In the mean time we have a PR that now encompasses multiple bug reports and feature requests. In a more perfect world, each of those should be their own issue, and we can then resolve them as we go.
> 
> So 1st step - I'm going to open a new issue to define a _minimal_ test suite, and notify the major players here. Possibly this is something @gmao-rreichle et al already have, but I'm guessing that their suite is larger than what I envision. Once the tool can pass this initial test suite, the PR should be merged. We can then mine this ticket for new issues/features and open corresponding tickets. (Happy to have others prioritize among them.)

This is a good idea. Thank you @tclune

sdrabenh:
@tclune @weiyuan-jiang defining a minimal test suite is a bit nebulous with such an all-encompassing tool like this. From my point of view, the new python script should be able to demonstrate comparable results to the pearl script for current (Icarus/Jason) datasets and resolutions. Regarding the functionality of the tool, it should be able to:

1. Run irrespective of what the working directory is
2. Resolve all relative or absolute paths correctly
3. Recognize whether it is running on login or interactive nodes and execute accordingly
4. Provide the correct/same default values for wemin, wemout, zoom, etc as the pearl script did
5. Accept command line arguments without requiring a yaml file so it doesn't break existing code [such as this](https://github.com/GEOS-ESM/GEOSgcm_App/blob/5526d3589e48ffe6cb74767f085e8d151bbd8302/gcm_run.j#L191)
6. Demonstrate consistent results whether running the script interactively, or with command line arguments, or using a yaml file

From an AGCM point of view, the most important atmospheric resolutions are C90, C180, C360, and C720. I would think that regridding between higher horizontal resolutions would better demonstrate whether the script is working correctly since they inherently contain more information. We frequently regrid between 72, 91, 137, and 181 levels. In fact, Bill's feature branch is aimed toward non 72-level resolutions, thus all those experiments. Lastly, it is common to regrid between the 4 data ocean resolutions based on whether we are running an OPS-like experiment, EMIPs, AMIPs, etc. Again, just seeing a few cases here demonstrating 0-diffness/equivalency with the pearl script would go a long way in establishing confidence this script is working correctly.

My wish list for this script (after meeting minimal requirements) would be to:
1. Automatically determine as much information about the input datasets as possible and pre-fill the defaults. In an ideal world the user shouldn't have to know anything about the input dataset, although I'm sure there are limitations to this.
2. Output the *.CMD text file the pearl script did. This was helpful.
3. When running interactively, provide the operation summary for the user to okay just like the pearl script did
4. Add back the coupled ocean resolutions - @sanAkel thoughts here?

sdrabenh:
@weiyuan-jiang I just tried your python script again and noticed you addressed several issues I mentioned above. Thanks! Improvements already. However, it is still crashing running the same test as yesterday. It got farther but it looks like another missing file issue. Please see my full log here:
`/gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python/output.log`

tclune:
@sdrabenh - moving your comments above to the issue rather than the PR

weiyuan-jiang:
@sdrabenh 
I addressed most of your comments, except
1) Command line. I will leave this feature for the future. It seems to me yaml config  file can be treated as command line inout.
2) ocean coupled model. I am not sure of the logic. The coupled model would show up if MOM5 or MOM6 is chosen. Is that right ? @sanAkel

sanAkel:
> 2. ocean coupled model. I am not sure of the logic. The coupled model would show up if MOM5 or MOM6 is chosen. Is that right ? 

that's right @weiyuan-jiang

sanAkel:
Some of you are aware ... but in case your memory needs to be refreshed!

We have been working (@rtodling and I) to add a way to use `atmos-only` restarts in `coupled` mode. @mathomp4 kindly made this possible. What that meant is that following option to re-grid was added:
- input: `data` _ocean_
- output: `MOM 5` or `MOM 6` ocean

This feature **must** be preserved. 

@weiyuan-jiang please note. Thanks!

weiyuan-jiang:
> 

@sanAkel Thanks for the confirmation.  I think regrid_py preserves that function.

sdrabenh:
> @weiyuan-jiang I just tried your python script again and noticed you addressed several issues I mentioned above. Thanks! Improvements already. However, it is still crashing running the same test as yesterday. It got farther but it looks like another missing file issue. Please see my full log here: `/gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python/output.log`

@weiyuan-jiang I just tried this exact same test case again this morning (3rd time) with your latest updates and I am getting the same error from `/discover/nobackup/projects/gmao/g6dev/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python.try3/output.log`

```
228-/gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/mk_LakeLandiceSaltRestarts.x /gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python//OutData//CF0360x6C_CF0360x6C-Pfafstetter.til /gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python//InData//CF0180x6C_DE1440xPE0720-Pfafstetter.til InData/Jason-3_4_NL_REAMIP_MERRA2_C180.openwater_internal_rst.e19850509_21z.Jason-3_4.Icarus-NLv3_MERRA-2_CF0180x6C_DE1440xPE0720.nc4 0 8
229-Traceback (most recent call last):
230-  File "/discover/nobackup/projects/gmao/g6dev/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/regrid.py", line 88, in <module>
231-    main(sys.argv[1:])
232-  File "/discover/nobackup/projects/gmao/g6dev/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/regrid.py", line 77, in main
233-    lls.regrid()
234-  File "/gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/regrid_lake_landice_saltwater.py", line 105, in regrid
235-    subprocess.call(shlex.split(cmd))
236-  File "/usr/local/other/python/GEOSpyD/4.10.3_py3.9/2022-01-14/lib/python3.9/subprocess.py", line 349, in call
237-    with Popen(*popenargs, **kwargs) as p:
238-  File "/usr/local/other/python/GEOSpyD/4.10.3_py3.9/2022-01-14/lib/python3.9/subprocess.py", line 951, in __init__
239-    self._execute_child(args, executable, preexec_fn, close_fds,
240-  File "/usr/local/other/python/GEOSpyD/4.10.3_py3.9/2022-01-14/lib/python3.9/subprocess.py", line 1821, in _execute_child
241-    raise child_exception_type(errno_num, err_msg, err_filename)
242:FileNotFoundError: [Errno 2] No such file or directory: '/gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/mk_LakeLandiceSaltRestarts.x'
```

Please use similar inputs and let me know when you can successfully run this case without errors executing `$GEOSUTIL/bin/regrid.py` from any arbitrary directory. I will wait to do further testing until I hear from you this case can be successfully run. Also, it looks like the interactive node issue was halfway-addressed based on my output log lines 191-194:

```
190-/bin/ln -s topo_DYN_ave_360x2160.data topo_dynave.data
191:interactive mode
192-
193-
194-You should have at least 96 cores. Now you only have 1 cores 
195-./regrid_upper.j  1>/gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python//regrid_upper_log  2>&1
```

weiyuan-jiang:
@sdrabenh I am surprised that there is no such file  /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/mk_LakeLandiceSaltRestarts.x . I have no access to your folder so I cannot figure that out. Maby different GEOSgcm_GridComp branch ?
In your interactive mode, can you post the result from the command?  " env | grep SLURM_  " 
I am wondering you may acquire the nodes in different way.

weiyuan-jiang:
@sdrabenh I actually had no problem running the tool in all locations and by interactive mode or by sbatch mode.  But I may have different setups from yours.

sdrabenh:
> @sdrabenh I am surprised that there is no such file /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/mk_LakeLandiceSaltRestarts.x . I have no access to your folder so I cannot figure that out. 

That is very strange. Everyone should have read/execute access to my directories:
```
drwxr-xr-x 28 sdrabenh g0620   32768 Apr 19 11:10 /gpfsm/dnb78s1/projects/p18/sdrabenh
drwxr-xr-x 42 sdrabenh g0620   32768 May  4 09:26 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models
drwxr-xr-x 25 sdrabenh g0620   32768 Jun  8 21:50 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs
drwxr-xr-x 13 sdrabenh g0620   32768 Jun  9 09:48 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py
drwxr-xr-x 13 sdrabenh g0620   32768 Jun  9 09:54 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install
drwxr-xr-x  8 sdrabenh g0620   32768 Jun  9 09:54 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin
-rwxr-xr-x  1 sdrabenh g0620 1954944 Jun  8 10:37 /gpfsm/dnb78s1/projects/p18/sdrabenh/Models/GIT_PRs/PRs_regrid.py/install/bin/mk_LakeLandiceSaltRestarts
```
and
```
drwxr-xr-x 28 sdrabenh g0620 32768 Apr 19 11:10 /gpfsm/dnb78s1/projects/p18/sdrabenh
drwxr-xr-x 11 sdrabenh g0620   512 Jun  7 13:31 /gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps
drwxr-xr-x  7 sdrabenh g0620 32768 Jun  9 10:38 /gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py
drwxr-xr-x  5 sdrabenh g0620 32768 Jun  9 16:11 /gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python.try3

```

> Maby different GEOSgcm_GridComp branch ? In your interactive mode, can you post the result from the command? " env | grep SLURM_ " I am wondering you may acquire the nodes in different way.

I prepended my environment to `/gpfsm/dnb78s1/projects/p18/sdrabenh/test_exps/PRs_regrid.py/test1_out_C360_cs_python.try3/output.log`

weiyuan-jiang:
@sdrabenh , Ok, I know what happened. You have not updated GEOSgcm_GridComp. So you did not see .x  executables. Here are my example of interactive run and sbatch run. They both run successfully. 
/discover/nobackup/wjiang/f522_interactive
/discover/nobackup/wjiang/f522_out

You got warning message  in interactive mode because you don't have SLURM_NTASKS env variable.  I guess you don't specify --ntasks when acquiring the resource. That is ok, I can add another way to calculate the ntasks.

sdrabenh:
> @sdrabenh , Ok, I know what happened. You have not updated GEOSgcm_GridComp. So you did not see .x executables. Here are my example of interactive run and sbatch run. They both run successfully. /discover/nobackup/wjiang/f522_interactive /discover/nobackup/wjiang/f522_out

Maybe I pulled your branch too early before your fix. I will try again.

> 
> You got warning message in interactive mode because you don't have SLURM_NTASKS env variable. I guess you don't specify --ntasks when acquiring the resource. That is ok, I can add another way to calculate the ntasks.

That is correct. Specifying ntasks is not necessary. In my case I asked for nodes instead. Glad you discovered that.

sdrabenh:
> I fixed the line
> 
> https://github.com/GEOS-ESM/GMAO_Shared/blob/c114d064cb0e0d6de93c7c4268300161610b791f/GEOS_Util/post/regrid_py/regrid_params.py#L536
> 
> @sdrabenh , I am sorry to put the debugging burden on you and thanks a lot for helping me to make this script more robust.

Great thanks. Also, in my random test case #3 not all restarts were produced, but there were no obvious errors in the output log:
`/discover/nobackup/projects/gmao/g6dev/sdrabenh/test_exps/PRs_regrid.py/test3_out_C90_o1_python/output.log`

weiyuan-jiang:
@sdrabenh , Sorry again for the bugs.  I made more changes ( please be aware of the changes of script's names)

1) change regrid.py to remap_restart.py
2)change the default zoom value
3)add support for restarts' file names without time stamp

To not waste your time, would you please give me some sets of results from regrid.pl ? I will do the tests and debugging myself.  Only after I passes all those tests, I would ask your help again. 
@biljanaorescanin

GEOS_Util/post/remap_restart/remap_questions.py
```diff
@@ -0,0 +1,415 @@
...
+        {
+            "type": "path",
+            "name": "rst_dir",
+            "message": "Enter the directory containing restart files to be remapped:",
+            "when": lambda x: not x['MERRA-2'],
+        },
+        {
+            "type": "path",
+            "name": "rst_dir",
+            "message": "Enter a directory to which the archived MERRA-2 archive files can be copied: ",
```
gmao-rreichle, line 49: For this question, can we have python create a default directory that the user can simply accept?
Something like:
`"default": '/discover/nobackup/$USER/tmpM2restarts/'`
(The syntax is just a wild guess and almost certainly won't work as is.)
My hope is that python will print the above directory to the screen (with "$USER" resolved properly) so that the user can just hit return to accept the suggestions and move on.  If the dir does not exist, the script should create it.

gmao-rreichle:
@weiyuan-jiang:  I tried to clarify the questions a little bit more.  Let me know if I messed things up.  Here's my commit: https://github.com/GEOS-ESM/GMAO_Shared/pull/238/commits/dd2c48f21b8694aa47f6138f7ea0a168e0937b4b
Also, see my [comment](https://github.com/GEOS-ESM/GMAO_Shared/pull/238#discussion_r911964926) about adding a default to the question about the M2 directory.

sanAkel:
Hi @weiyuan-jiang, @tclune are there are issues in fixing the errors that @sdrabenh pointed to ⬆️. Please [see](https://github.com/GEOS-ESM/GMAO_Shared/pull/238#issuecomment-1155139229). Thanks!

GEOS_Util/post/remap_restart/remap_questions.py
tclune: I would prefer that defaults not reference absolute paths, as it erodes portability.   But we could have a notion of $NOBACKUP that must be specified for any platform (somewhere) and then relative paths from that.

tclune:
@sanAkel Not sure where things stand with regard to @sdrabenh issues, but we certainly want them all fixed before wasting more of his time.   Last I checked, @weiyuan-jiang had fixed several of the issues.  He'll have to update us here.

sanAkel:
@yvikhlya when time permits, would you please look into issues with MOM5. Thanks!



> * MOM5 case (if I pick MERRA2 CS to go to MOM5 720x410) : this case will produce .bin catch restart with regrid.pl but not with rebase_restarts.py, Weiyuan said we don’t support catch restarts in .bin format anymore. But it is zero diff between the files.

weiyuan-jiang:
After all the tests are settled, the PR #272 would re-order the questions

sdrabenh:
@weiyuan-jiang have you started from a clean build recently? I tested this yesterday and the executable names need to be consistent or else this script fails. It is looking for `mk_LakeLandiceSaltRestarts.x` but only `mk_LakeLandiceSaltRestarts` exists. It is looking for `mk_catchANDcnRestarts.x` but only `mk_CatchCNRestarts` and `mk_CatchRestarts` exist. Please be sure all your names are consistent and retest from a clean build.

biljanaorescanin:
Talking to @sdrabenh it seems this is a build issue. When I run cases, I run them from `install/bin` directory and for me all @sdrabenh failed cases were successful, no fails or errors. 
For example:
`/discover/nobackup/borescan/brisi/test_regrid/try_22/test3_out_C90_o1_python.2`
When he runs he uses `$GEOSUTIL/bin/remap_restarts.py` and somehow has issue with renamed files,  some`.x` are missing.

gmao-rreichle:
@sdrabenh, @biljanaorescanin, @mathomp4:  
@weiyuan-jiang added the `.x` extension to the program executable names for consistency with the rest of GEOS-ESM.  I understand from @biljanaorescanin that the sym-links that @weiyuan-jiang introduced don't work for @sdrabenh when he runs regrid.pl from a place other than install/bin.  If links don't work, we could just keep two copies of the executables, one with .x and one without .x.  The copy without .x would then be discontinued when we discontinue the perl package, presumably later this year.  If this approach is ok, @biljanaorescanin can look into replacing the sym-links with a copy.

sdrabenh:
> @sdrabenh, @biljanaorescanin, @mathomp4: @weiyuan-jiang added the `.x` extension to the program executable names for consistency with the rest of GEOS-ESM. I understand from @biljanaorescanin that the sym-links that @weiyuan-jiang introduced don't work for @sdrabenh when he runs regrid.pl from a place other than install/bin. If links don't work, we could just keep two copies of the executables, one with .x and one without .x. The copy without .x would then be discontinued when we discontinue the perl package, presumably later this year. If this approach is ok, @biljanaorescanin can look into replacing the sym-links with a copy.

@biljanaorescanin, @weiyuan-jiang, @gmao-rreichle, just to clarify, my "clean" build from checking out this branch on top of main produces an `install/bin` directory with a mix of executables:
```
 ls -1 mk_*
mk_CatchCNRestarts
mk_CatchRestarts
mk_CICERestart
mk_GEOSldasRestarts
mk_LakeLandiceSaltRestarts
mk_Restarts
mk_RouteRestarts
mk_runofftbl.x
```
Most do not have the `*.x` extension and no sym links exist. When I run `$GEOSUTIL/bin/remap_restarts.py` interactively from a separate working directory the scripts fail because they are looking for 2 non-existent files `.../install/bin/mk_LakeLandiceSaltRestarts.x` and `.../install/bin/mk_catchANDcnRestarts.x`

biljanaorescanin:
@sdrabenh but isn't that the problem? My build has that, those sym links.  And all 4 tests are zero diff to pearl. 
Screen Shot 2022-07-13 at 10 19 34 PM

sdrabenh:
> @sdrabenh but isn't that the problem? My build has that, those sym links.

Correct. Do your sym links exist after building the model or after running `remap_restarts.py`? Either way, something does not appear to be working correctly and the question is why.

sdrabenh (approved):

yvikhlya (approved):

mathomp4 (approved):
Approve for CMake!

PR #240, opened 2022-02-03
WIP: Changes needed for NAG Compiler

mathomp4:
This PR tracks changes needed to compile GMAO_Shared with the NAG Compiler. NAG is *very* strict and will not allow the use of Fortran extensions that are endemic to GEOS. 

---

The first category are procedures that are now part of the Fortran Standard that used to be extensions in some compilers:

- `iargc()` → `command_argument_count()`
- `getarg()` → `get_command_argument()`
- `getenv()` → `get_environment_variable()`
- `system()` → `execute_command_line()`
- `call exit(0)` → `stop`
- `call exit(N)` → `error stop N`

NAG also doesn't support (without flags that need to be used *everywhere* as @tclune can elucidate on) the use of `real*4` or `real(kind=4)`. The reason is that NAG does not use 4 for 32-bit real and 8 for 64-bit, instead they use 1 and 2. (The Standard does not specify what "kind" a 32-bit real must be, only that there must be one.) So, many changes are of the type:

- `real*4` → `real(REAL32)`
- `real(kind=8)` → `real(kind=REAL64)`


rtodling (dismissed):
These are fine by me. Branch might need update though.

mathomp4:
> These are fine by me. Branch might need update though.

Yeah. I need to revisit this. Files got moved long ago...

I'll put it on the list!

weiyuan-jiang:
I made a new commit but I didn't do the zero test @mathomp4

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ if ( NOT HERMES_LIGHT )
      GMAO_mpeu
      GMAO_pilgrim
      GMAO_etc
-     GEOS_Util
+     @GEOS_Util
```
weiyuan-jiang, line 14: It seems GEOS_Util has been deleted. But this @GEOS_Util should be here?

weiyuan-jiang:
@mathomp4 , The fix is not working any more. I think it needs more fixes.

weiyuan-jiang:
@mathomp4 Never mind. It needs the help from [ESMA_Cmake](https://github.com/GEOS-ESM/ESMA_cmake/pull/541)

PR #259, opened 2022-04-11
Use the monthly attribute by default

sdrabenh:
Change the default output of most collections to use MAPL's monthly mean and compression. Turn off automatic archiving to mass storage. Changes enable compatibility with plots package.


mathomp4 (approved):
Trust in Scott. Also: Don't forget CHANGELOG!

mathomp4 (approved):

PR #261, opened 2022-05-12
Allow for non-quad-precision systems

mathomp4:
In testing GEOS on an M1 MacBook, it was found that GNU on M1 does not support quad-precision (aka REAL128). Stoch currently uses this in its kinds. So, what this PR does is back down quad precision code to double precision if not supported. If it is supported, then nothing changes.

I'll add @elakkraoui to the reviewers as, well, I think of her when I think of this code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,13 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


mathomp4:
I am converting this to draft for now. It's possible the latest 11.3 branch of GCC for M1 actually does have 128-bit support. Or, at least, has enough support for this not to matter...yet. (PGI and NAG I think don't support quad at all, so we might need to fix this eventually.)

Issue #263, opened 2022-05-27
GEOSachem restart improperly named in regrid.pl


bena-nasa:
A user reported this:
"I’ve checked out the latest main branch of the model and not that in regrid.pl the achem restart is incorrectly named.
 
Please change the name in the perl script from “geosachem_internal_rst” to “achem_internal_rst” going forward. It is correctly labeled in the AGCM.rc templates."

PR #264, opened 2022-05-27
geocachem ==> achem name changed

weiyuan-jiang:
Address #263

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,42 +1,20 @@
 version: 2.1
 
 orbs:
-  circleci-tools: geos-esm/circleci-tools@0.10.0
+  ci: geos-esm/circleci-tools@1
 
 workflows:
   build-test:
     jobs:
-      - build-GEOSgcm:
+      # Build GEOSgcm
+      - ci/build:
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
I have a concern. After this merge, the regrid.pl is not going to support the old name "geosachem" any more. Would that be OK ? @bena-nasa

mathomp4:
> I have a concern. After this merge, the regrid.pl is not going to support the old name "geosachem" any more. Would that be OK ? @bena-nasa

@weiyuan-jiang Well, we can go with the "advice" we've always had: Use the `regrid.pl` from the model tag you want to run. 

So, if you get a 10.22 GEOS, you'd expect it to look for `achem_internal_rst`. If it's an older one (GOCART1 days?) then `regrid.pl` would look for the "appropriate" `geosachem_internal_rst` file. I guess we sort of "broke" that in recent times. 

If someone wants to use the latest `regrid.pl` with some old restarts, well, we might need to advise them to edit the perl script. I suppose you could put a comment in the code mentioning that this restart's name changed?

mathomp4:
I pushed an update to try and fix the CI which is oddly failing.

sdrabenh (approved):

PR #266, opened 2022-06-13
rename regrid to remap

weiyuan-jiang:
It is not accurate to use "regrid".  Maybe remap is a better name.


tclune (commented):
cmake changes ok

Issue #268, opened 2022-06-23
Stress across tripole fold not being updated properly in CICE


zhaobin74:
In CICE EVP dynamic routine, there is a block of code supposed to force symmetry of ice internal stresses across the tripolar fold/seam. However, this logic only runs after EVP subcycling steps are finished while stresses are being updated at every substep. This is the cause of ice getting huge deformation and ridging rate and in one configuration crashes the model.

zhaobin74:
Not relevant any more. The issue was fixed by grid file changes.

PR #269, opened 2022-06-23
Bugfix/zhaobin74/#268 force symmetry tripole in subcycle

zhaobin74:
This PR fixed issue #268.

- The block of code enforcing stress symmetry across the tripolar fold is moved inside the EVP subcycling loop.
- This change fixed the spurious ice ridging problem as seen with some MOM6 OM4 grids.
- This PR is 0-diff in uncoupled mode and fixes a bug in coupled mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - More updates to CMake to more canonical CMake style (NetCDF, ESMF, etc.). These were missed in previous go-arounds as they are only built with ADAS. (Requires ESMA_cmake v3.15.1)
 
+- Fixed a bug in CICE4 EVP dynamic code to force symmetry across the fold in every subcycle step
+
 ### Removed
 
 ## [1.5.4] - 2022-05-03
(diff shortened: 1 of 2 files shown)


sanAkel (approved):
We need this fix. Thanks @zhaobin74

zhaobin74:
> I've added the "0-diff" label as our label enforcer only understands the simple ones. I'll have to look at adding the "uncoupled" labels to it as well

Thanks, Matt. Yes, we need a more clear label system as to 0-diff relative to uncouple vs coupled.

sanAkel:
> I've added the "0-diff" label as our label enforcer only understands the simple ones. I'll have to look at adding the "uncoupled" labels to it as well

@mathomp4 Maybe easier to read _smaller_ labels such as "AGCM", "AOGCM", ...

zhaobin74:
It turned out that the sea ice problem along the zipper is due to a grid file issue. This PR is to be retracted.

sanAkel:
@zhaobin74 , @yvikhlya,

I am feeling more inclined to write out a `grid_spec.nc` _style_ file (as with MOM5) that has ALL the bits from `ocean_static.nc`. Let's discuss this via MS Teams. Thanks!

PR #281, opened 2022-08-17
Add S2Sv3 to MOM6 test. Add YAML validator

mathomp4:
This PR adds an S2Sv3-MOM5 to C12-MOM6 test case to the remapping tests. It also has some fixes from @weiyuan-jiang while debugging other issues I found/created. 😄 

It also adds a YAML validator and cleans up the existing YAML files to pass.


mathomp4:
Converting this to draft just until @weiyuan-jiang is satisified.

PR #299, opened 2022-10-28
Feature/bmauer/grid agnostic time ave

bena-nasa:
I have done a huge refactoring of time_ave.x to create new time averaging utility that will support any grid the GEOS model can output via History. At this time, this set includes lat-lon, cubed-sphere, and tripolar.

Rather than refactoring the existing time_ave.F (it could not stay an .F anyway), I created a new executable and source file which I called time_ave_util.F90 which compiles to time_ave_util.x

Note with this new code MPI is not optional. It can still be run on 1 processor but must be built with MPI. In addition to use cubed-sphere files, it must be run on a minimum of 6 or multiple of 6.

Finally I added a new option so that if you want the averaging to occur on a different grid that the input grid of the files you can specify a grid name via new -ogrid argument. For the convention I use for the grid name see the wiki page about my Regrid_Util.x in MAPL

Note this does require the MAPL v2.29.0 version we just released. It will not build without it.

Note also that I did change time_ave.F in a few places, this is so the debug build would run and I could test side by side with the new executable in the debug build which I develop in. apparently no one every tested it with the debug build :(


tclune (commented):
Given the comments under "discussion", I am surprised that "time_ave.F" was touched at all.  Granted the changes there were small, but were they intended?

I suspect the new program can be simplified and/or clarified,   but am guessing most of the sins on that front are from duplicated the older code.  (E.g., the indentation is a giveaway.)

My major concern (and why I'm not clicking "Approve") is that the dependency on a recent MAPL may cause issues for @rtodling .     Pinging @mathomp4 to remind me what we can check about the DAS builds with the current CI.

PR #310, opened 2022-12-29
Feature/bweir/#305 codas

briardew:
Updates to GMAO_ods to be able to process trace gas obs.


CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## feature/bweir/#305-codas
```
mathomp4, line 30: @briardew You can remove this bit. We only need the entry after `### Added`

mathomp4:
Oh. Actually, worse than that. @briardew Did you branch off of something older than `main`? If so, we'll have to make a release branch for you since it looks like you are changing code from long ago.

briardew:
Well yes and no. This is a branch off of what's used in the GEOSadas develop branch, which uses the v1.4.10.2 tag for GMAO_Shared. Everything in the GMAO_ods directory is identical to that and the main branch of GMAO_Shared. Problem is, if I checkout the main branch of GMAO_Shared in the develop branch of GEOSadas, it doesn't compile. I suspect because they want different things. So I branched off of the thing that works in GEOSadas. Happy to take any suggestions on how to proceed.

PR #326, opened 2023-06-05
Brought  LANL_Shared/CICE4/source/ice_therm_vertical.F90 from

atrayano:
This resets the ice to default profile under some error conditions. Should be zero difff. It affects the ice only when otherwise the code would have crashed. Same mods are already in use in S2Sv3

### DIFF
--- a/LANL_Shared/CICE4/source/ice_therm_vertical.F90
+++ b/LANL_Shared/CICE4/source/ice_therm_vertical.F90
@@ -498,7 +498,31 @@
                                   l_stop,                     &
                                   istop,        jstop)
 
-      if (l_stop) return
+      if (l_stop) then
+
+         l_stop= .false.
+          
+         call fixed_vertical_profile (nx_block,     ny_block, &
+                                  my_task,      istep1,       &
+                                  icells,                     &
+                                  indxi,        indxj,        &
+                                  aicen(:,:),                 &
+                                  vicen(:,:),   vsnon(:,:),   &
+                                  Tsfcn(:,:),                 &
+                                  eicen(:,:,:), esnon(:,:,:), &
+                                  hin,          hilyr,        &
+                                  hsn,          hslyr,        &
+                                  qin,          Tin,          &
+                                  qsn,          Tsn,          &
+                                  Tsf,          einit,        &
+#ifdef GEOS
+                                  Tbot,                       &
+                                  tlat, tlon,                 &
+#endif
+                                  l_stop,                     &
+                                  istop,        jstop)
+         if (l_stop) return   
+      endif
 
       do ij = 1, icells
          ! Save initial ice and snow thickness (for fresh and fsalt)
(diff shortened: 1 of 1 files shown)


sanAkel:
@zhaobin74 can you comment/check to make sure answers reproduce? Thanks!

zhaobin74:
@atrayano, I am not sure if this is zero-diff. If an error condition occurs (for whatever reason), this fix will allow model to continue while the model stops without this step. 

Normally, if the error occurs, (l_stop is true), we should let the model blow up as that indicates some problems somewhere. if we just fix the profile, that will hide the problem.

Is this done mainly for MITgcm? If so, it should be MIT only.

atrayano:
Yes, this is for MITgcm. We already used this code in the DYAMOND runs, and it was already on a dyamond related branch of mine. However, it is already part of S2Sv3, and @amold suggested that this should be merged in. My definition of "zero-diff" is when the code runs, this change is not active, and should not affect the results. It is only a workaround to prevent code crashes

zhaobin74:
@atrayano, I don't recall this is added in S2SV3. I certainly didn't do it. Plus, S2SV3 never needs it. I am afraid that adding this will hide the true problem if there is one. My suggestion is to only execute this part when coupled to MITgcm. For mom5/6, I 'd like model to blow up so we can figure out the real problem. 


> Yes, this is for MITgcm. We already used this code in the DYAMOND runs, and it was already on a dyamond related branch of mine. However, it is already part of S2Sv3, and @amold suggested that this should be merged in. My definition of "zero-diff" is when the code runs, this change is not active, and should not affect the results. It is only a workaround to prevent code crashes

atrayano:
@zhaobin74 , ok I will protect the changes to be in place only when MITgcm is active. Otherwise the code would crash when the problem occurs

zhaobin74:
@atrayano, that will be great. Thanks.

> @zhaobin74 , ok I will protect the changes to be in place only when MITgcm is active. Otherwise the code would crash when the problem occurs

atrayano:
Done. The protection is now active only when we compile MIGgcm (i.e. cmake is executed with "-DBUILD_MIT_OCEAN"

zhaobin74:
@atrayano, could you point out where the protection (MITgcm only) is enforced? Thanks

> Done. The protection is now active only when we compile MIGgcm (i.e. cmake is executed with "-DBUILD_MIT_OCEAN"

mathomp4:
@atrayano I updated the changelog. Something was looking weird in Markdown

atrayano:
Thanks for your message. I did not realize that the  "git push" had failed. All good now!

Here is the protection[
](https://github.com/GEOS-ESM/GMAO_Shared/blob/0eb4ee49bc5c9a6570725bfa2140a9f51d430063/LANL_Shared/CICE4/source/ice_therm_vertical.F90#L501)

And there is a change in the cmake file to pass that flag

atrayano:
@mathomp4 , I had to merge, so I might have trashed accidentally your CHANGELOG changes. Could you take another look?

zhaobin74:
Now I see it. It looks good. Thanks for making the change.

> Thanks for your message. I did not realize that the "git push" had failed. All good now!
> 
> Here is the protection[ ](https://github.com/GEOS-ESM/GMAO_Shared/blob/0eb4ee49bc5c9a6570725bfa2140a9f51d430063/LANL_Shared/CICE4/source/ice_therm_vertical.F90#L501)
> 
> And there is a change in the cmake file to pass that flag

sdrabenh (dismissed):

zhaobin74 (approved):

PR #331, opened 2023-09-15
sync obs sys w/ fp; revise util progs

rtodling:
This syncs ops-sys w/ that used in FP as of 14 Sep 2023. 

It also revises two utility programs:

1) dyn2dyn - allows for user specified IM x JM when regrinding
2) dyn_hydro - somehow the original for this seems to have been lost in the CVS migration! This program adds the hydrometeors to bkg.eta files that don't have them (such as those coming from MERRA-2).

None of the changes affect anything - the obs-sys change only affects experiments running after 1 Sep 2023 (none exist at this time).


rtodling:
BTW: If someone w/ the power to version-tag this can give this a tag and release version, I'd appreciate. I need a version number to put it in the components.yaml - not sure I have the version-power for this repo ... thanks.

mathomp4 (approved):

PR #348, opened 2024-05-28
vectorize the operation in helfsurfce and other subroutines

weiyuan-jiang:
Try to vectorize the subroutine and eliminate some unnecessary calculation to improve the performance of  the subroutine helfsurface

PR #350, opened 2024-06-04
Update CI to use Intel 2024.1 and Intel MPI 2021.12

mathomp4:
@weiyuan-jiang Says that Intel 2021.6 has an issue with his updated code in #348 . So this is a test to update the CI to use a newer Intel.


rtodling (approved):
Oh yes - by all means - who uses this?!

PR #356, opened 2024-08-15
gcc14: Add int to main in makdep.c

mathomp4:
Building GEOS with GCC 14 found an issue where `makdep.c` did not declare `main` as an `int`. GCC 14 no longer allows implicit int main it seems.


zhaobin74 (approved):
Long overdue. Thanks for the fix.

sdrabenh (approved):

PR #366, opened 2025-01-17
Build GMAO_ncdiag in GMAO_Shared with gfortran-13 

metdyn:
This is a note to @mathomp4:
I am only building  src/GMAO_Shared/GMAO_ncdiag with gfortran-13,  excluding all other directories in GMAO_Shared.
I find GNU compiler complains many small Fortran errors and I find the logic to fix them.
I have  in src/CMakeLists.txt
```
if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
    set(BIG_ENDIAN "-fconvert=swap")
endif()

add_subdirectory (GMAO_Shared)
```
as I find a build error that `BIG_ENDIAN` was not defined, but this did not occur when building more directories in GMAO_Shared for GNU (hence it is just a stand alone test note).

### DIFF
--- a/GMAO_ods/CMakeLists.txt
+++ b/GMAO_ods/CMakeLists.txt
@@ -116,7 +116,7 @@ separate_arguments(ODSMETA_ARGS NATIVE_COMMAND "sed -e 's/integer, parameter//g'
 add_custom_command (
    OUTPUT odsmeta.py
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-   SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/odsmeta.h
+   DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/odsmeta.h
    COMMAND grep 'integer, parameter' ${CMAKE_CURRENT_SOURCE_DIR}/odsmeta.h | ${ODSMETA_ARGS} | grep -v idsats > odsmeta.py
    )
 
--- a/GMAO_ncdiag/nc_diag_cat/nc_diag_cat.F90
+++ b/GMAO_ncdiag/nc_diag_cat/nc_diag_cat.F90
@@ -77,7 +77,7 @@ program nc_diag_cat
     call nc_diag_cat_metadata_pass
     call cpu_time(stop_time)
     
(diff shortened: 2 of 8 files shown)


GMAO_ods/CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ separate_arguments(ODSMETA_ARGS NATIVE_COMMAND "sed -e 's/integer, parameter//g'
 add_custom_command (
    OUTPUT odsmeta.py
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-   SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/odsmeta.h
+   DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/odsmeta.h
```
metdyn, line 119: I am not sure about this change

mathomp4:
@metdyn Huh. I just looked an in the [GNU Flags](https://github.com/GEOS-ESM/ESMA_cmake/blob/80f22ba2b6b2fc65a3b6eb59a1b2cf131093f546/compiler/flags/GNU_Fortran.cmake#L93-L94) we set in ESMA_cmake I see:

```cmake
#set (BIG_ENDIAN "-fconvert=swap") # This doesn't seem to work at the moment
#set (LITTLE_ENDIAN "") # Not sure
```

Most likely I didn't have any code to test with so I played it safe and turned it off? Not sure. The GNU docs say:

> -fconvert=conversion
> 
> Specify the representation of data for unformatted files. Valid values for conversion on most systems are: ‘native’, the default; ‘swap’, swap between big- and little-endian; ‘big-endian’, use big-endian representation for unformatted files; ‘little-endian’, use little-endian representation for unformatted files.

So I suppose `swap` does what we need, but maybe these should be set to `little-endian` and `big-endian`? (Though if we ever need to set a `LITTLE_ENDIAN` flag, we are running on the *wrong* system.)

metdyn:
@mathomp4, thanks for looking into this. 
I realized setting gfortan flags can reduce a lot of error messages when compiling legacy Fortran codes. I will check offline with these options. Thanks!

PR #370, opened 2025-02-03
In Perl_Config.pm: Fixed the perl left/right brace matching problem. …

gmao-jstassi:
…The right brace now needs to be escaped similar to the left brace.


saraqzhang:
@gmao-rreichle ,  this fix is used in the integrated setup of LADAS.

sdrabenh (dismissed):

PR #381, opened 2025-05-29
Bugfix/dao ops/sles15 fix

rlucches:
r_dist is used in post-processing utilities in the ops implementation of FP.  The version in the tag still had a construct that worked under SLES-12, but fails under SLES-15.   This has been fixed.


mathomp4:
Hmmm. I don't think this is meant to go into `GEOS-FP`. You didn't change 1000 files.

From the git logs this seems to be based off of the `g5.42.2` from @rtodling.

I think we need to do a couple things. First, @rtodling should tag off of this branch. But then this PR should go to `main`, so we get the changes into the mainline of GMAO_Shared development.

mathomp4:
I just changed the base to be to `main`. Now I see four commits:

https://github.com/GEOS-ESM/GMAO_Shared/pull/381/commits

two from @rtodling and two from @rlucches

rlucches:
I changed the name of this PR to something sensible.  I have no idea how changes to hermes and transf ended up in this PR.  I did not change anything there.  The only change I made was to r_dist and perhaps the CHANGELOG.  It looks like those other changes are from Ricardo, but I am not sure how they got lumped into #381.

mathomp4:
@rlucches The reason why the @rtodling changes are in this is because you started from `g5.42.2` (which is what the ADAS was using). There are changes in that tag that are not yet in `main` here.

So, when you make a PR from your branch to `main`, it's all the changes not yet in `main`.

PR #398, opened 2025-12-09
Fix for GNU

mmanyin:
There was a runtime error when compiled with GNU and -fsanitize=leak.
The fix was to recast all vector declarations such as  REAL ABC(1)  to be instead REAL ABC(*) in  sphere.F .
Verified zero-diff for GCM when using the standard ifort compilation.
I recognize that this section of code may be used in several contexts; what other tests are required?


mathomp4:
Mentioning @sdrabenh and @wmputman as they probably go near `sphere.F` more than any of us.

sdrabenh (approved):

PR #412, opened 2026-04-01
Restrict GMAO_Shared to HERMES_LIGHT on release/MAPL-v3

tclune:
## Summary

- Sets `HERMES_LIGHT=ON` and builds only `GMAO_mpeu`, `GMAO_hermes` (light), and `GEOS_Shared` — the minimum needed by `FVdycoreCubed_GridComp`
- Extends the `HERMES_LIGHT` source list in `GMAO_hermes` to include `m_topo_remap.F90`, `m_dyn.f90`, and `m_const.f90` (needed via `use` statements in `FVdycoreCubed_GridComp`)
- Comments out the full `GMAO_Shared` build (pilgrim, transf, gfio, stoch, etc.) and `NCEP_Shared` — restore when all components are ported to MAPL3

## Context

Part of the MAPL3 flip work tracked in GEOS-ESM/GEOSgcm_GridComp#1359. Only components already ported to MAPL3 are compiled on `release/MAPL-v3`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,55 @@
 esma_check_if_debug()
 
-# HERMES_LIGHT builds only m_set_eta.F90 from GMAO_hermes, which is
-# required by FVdycoreCubed_GridComp on release/MAPL-v3.
-option(HERMES_LIGHT "Build light-weighted GMAO_shared library" ON)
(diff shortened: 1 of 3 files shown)


mathomp4:
I think the failing CI might be due to a race condition? As some of the `feature/deactivate-nonported-mapl3-components` branches weren't yet made? Not sure. I'm doing a full build of all the branches now.

PR #448, opened 2026-06-08
Fix for ods cmake

mathomp4:
This PR fixes a small CMake issue in GMAO_ods. The old code threw a warning and this is the better way to do it.


mathomp4:
Build failure? What? I mean, the CMake code I changed isn't even used in GEOSgcm!

sdrabenh (approved):

